PR #4, opened 2016-01-25
Spec dir structure matches model dir structure

ehannes:
When model directory structure was changed, corresponding structure
in spec directory was not updated.


ehannes:
I realize now that this pull request comes from my master branch. A better approach would be to keep my branches separate, create pull requests to this repo and once they are merged I can update my master branch.

d-Pixie:
Yeah, you can get into trouble with multiple PRs open from the same branch if one of the earlier ones are rejected or requires rework later. I'll try to keep on top of them though, so you should be fine. But I agree that best practice is using branches, one for each PR.

PR #5, opened 2016-01-26
Contribute

ehannes:
README.md updated with information about how to add original repo as remote server.

I now realize that e2d2ff3 should not be included in this PR. The problem was that I merged that commit to my master branch before I created a PR to this repo. Therefore, #4 should be merged before this PR can be merged. Sorry.


d-Pixie:
See, good example of why to use branches locally :D

d-Pixie:
Nice docs on contributing. I like it.

PR #8, opened 2016-01-27
Row

ehannes:
This is a temporary RP, not intended to be merged just yet. There are still models missing for the `Row` entity. I just like your opinion on the code (especially the `validator_spec` for `Row`).

### DIFF
--- a/lib/fortnox/api/models/row.rb
+++ b/lib/fortnox/api/models/row.rb
@@ -5,6 +5,9 @@ module Fortnox
     module Model
       class Row < Base
 
+        DISCOUNT_TYPES = %w(AMOUNT PERCENT)
+        HOUSE_WORK_TYPES = %w(CONSTRUCTION ELECTRICITY GLASSMETALWORK GROUNDDRAINAGEWORK MASONRY PAINTINGWALLPAPERING HVAC CLEANING TEXTILECLOTHING COOKING SNOWPLOWING GARDENING BABYSITTING OTHERCARE TUTORING OTHERCOSTS)
+
         #AccountNumber Account number. 4 digits
         attribute :account_number, Integer
 
@@ -61,6 +64,16 @@ module Fortnox
 
         #VAT VAT percentage of the row.
         attribute :vat, Integer
+
+        def discount_type=( raw_discount_type )
+          discount_type = raw_discount_type.upcase
+          super discount_type if DISCOUNT_TYPES.include? discount_type
+        end
+
+        def house_work_type=( raw_house_work_type )
+          house_work_type = raw_house_work_type.upcase
+          super house_work_type if HOUSE_WORK_TYPES.include? house_work_type
+        end
       end
     end
   end
--- a/lib/fortnox/api/models/row.rb
+++ b/lib/fortnox/api/models/row.rb
@@ -30,6 +30,8 @@ module Fortnox
         attribute :description, String
 
         # Discount amount. 12 digits (for amount) / 5 digits (for percent)
+        # TODO: Verify that we can send in more than 5 digits through the actual
+        # API for DiscountType PERCENT.
         attribute :discount, Float
 
         #DiscountType The type of discount used for the row.
--- a/lib/fortnox/api/models/row.rb
+++ b/lib/fortnox/api/models/row.rb
@@ -39,7 +39,7 @@ module Fortnox
         attribute :house_work_hours_to_report, Integer
 
         #HouseWorkType The type of house work.
-        attribute :HouseWorkType, String
+        attribute :house_work_type, String
 
         #Price Price per unit. 12 digits
         attribute :price, Float
--- a/lib/fortnox/api/models/row.rb
+++ b/lib/fortnox/api/models/row.rb
@@ -1,9 +1,10 @@
-require "fortnox/api/models/entity/base"
+require "fortnox/api/models/base"
 
 module Fortnox
   module API
-    module Entities
-      class Row < Fortnox::API::Entities::Base
+    module Model
+      class Row < Base
+
         #AccountNumber Account number. 4 digits
         attribute :account_number, Integer
 
--- a/lib/fortnox/api/models/row/entity.rb
+++ b/lib/fortnox/api/models/row/entity.rb
@@ -2,7 +2,7 @@ require "fortnox/api/models/entity/base"
 
 module Fortnox
   module API
-    module Enteties
+    module Entities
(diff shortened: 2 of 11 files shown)


ehannes:
For instance, I see that CodeClimate does not like this: ``subject { described_class.new( entity ) }``. "Space detected to the left of {.". Is that what you intended when changing the Rubocop configuration? Should I instead write ``subject{ described_class.new( entity )}``?

d-Pixie:
Yeah it's for all block syntax and should line up with using parentheses you write: `method( arg )` and in a block that is: `method{ block }`, basically ...

lib/fortnox/api/models/row/entity.rb
```diff
@@ -0,0 +1,66 @@
+require "fortnox/api/models/entity/base"
+
+module Fortnox
+  module API
+    module Entities
+      class Row < Fortnox::API::Entities::Base
```
d-Pixie: Does it need to inherit from `Base`? If it's only a value object it shouldn't have to. Look at it as a larger version of `Currency` or `CountryCode` ...

lib/fortnox/api/models/row/validator.rb
```diff
@@ -0,0 +1,19 @@
...
+    module Validators
+      class Row < Fortnox::API::Validators::Base
+
+        attribute :account_number, inclusion: (0..9999)
+        attribute :article_number, size: 1..50
+        attribute :delivered_quantity, inclusion: (0..99_999_999_999_999)
+        attribute :description, size: 1..50
+        attribute :discount, inclusion: (0..999_999_999_999)
+        attribute :house_work_hours_to_report, inclusion: (0..99_999)
+        attribute :price, inclusion: (0..999_999_999_999)
```
d-Pixie: Organise these by type: presence, size, inclusion. Separate the different types by a blank line.

You should also be using `validates` and not `attribute` since the updated use of Hanami.

lib/fortnox/api/models/row/entity.rb
ehannes: No, it doesn't. It cannot be updated or saved since it is not a Fortnox resource, so it should not be an entity at all. The question is then where to put this model. `Currency` and `CountryCode` is placed in `fortnox/api/attributes/`. I would say that `Row` is more than an attribute...

d-Pixie: Yeah, I agree. Keep it in models, but use the same basic principle as with the attribute classes.

ehannes: If I should keep it in models, what module should I use? The current module is a bit misleading since it's not an entity: `Fortnox::API::Entities`. Also, the class itself is named `entity.rb`. Maybe we could use `Fortnox::API::Models` and place it directly in at `lib/fortnox/api/models/row.rb`.

lib/fortnox/api/models/row.rb
```diff
@@ -0,0 +1,67 @@
...
+        attribute :discount_type, String
+
+        #HouseWork If the row is housework
+        attribute :house_work, Boolean
+
+        #HouseWorkHoursToReport Hours to be reported if the quantity of the row should not be used as hours. 5 digits
+        attribute :house_work_hours_to_report, Integer
+
+        #HouseWorkType The type of house work.
+        attribute :HouseWorkType, String
```
d-Pixie: What are the possible types?

ehannes: Good point! I'll look into it!

lib/fortnox/api/validators/row.rb
```diff
@@ -0,0 +1,20 @@
...
+      class Row < Fortnox::API::Validator::Base
+
+        validates :article_number, size: 1..50
+        validates :description, size: 1..50
+
+        validates :account_number, inclusion: (0..9999)
+        validates :delivered_quantity, inclusion: (0..99_999_999_999_999)
+        validates :discount, inclusion: (0..999_999_999_999)
+        validates :house_work_hours_to_report, inclusion: (0..99_999)
+        validates :price, inclusion: (0..999_999_999_999)
```
d-Pixie: Check the code for the size validation in Vanguard, it might use a comparator that allows us to convert all of these to size instead of inclusion. Inclusion is expensive :)

spec/fortnox/api/validators/row_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    context 'with invalid attributes' do
+      it_behaves_like 'invalid', :account_number, [-1, 12345]
+      it_behaves_like 'invalid', :article_number, ['a' * 51]
+      it_behaves_like 'invalid', :delivered_quantity, [-1, 123456789012345]
+      it_behaves_like 'invalid', :description, ['a' * 51]
+      it_behaves_like 'invalid', :discount, [-1, 1234567890123]
+      it_behaves_like 'invalid', :house_work_hours_to_report, [-1, 123456]
+      it_behaves_like 'invalid', :price, [-1, 1234567890123]
+    end
```
d-Pixie: Nice test :) I like.

spec/spec_helper.rb
```diff
@@ -6,6 +6,8 @@
 
 CodeClimate::TestReporter.start
 
+Dir[File.expand_path('spec/support/matchers/*.rb')].each{ |file| require file }
```
d-Pixie: Could create a meta file called `spec/support/matchers.rb` and include it. And keep the logic for loading the individual matchers in it...

spec/support/matchers/validators.rb
```diff
@@ -0,0 +1,30 @@
+RSpec::Matchers.define :validate_true do
```
d-Pixie: `:be_valid` instead of `:validate_true`

spec/support/matchers/validators.rb
```diff
@@ -0,0 +1,30 @@
...
+  expect_validate( true )
+end
+
+RSpec::Matchers.define :validate_false do
```
d-Pixie: `:be_invalid` instead of `:validate_false`

lib/fortnox/api/validators/row.rb
ehannes: Actually, I have tried size but without any success. It appears as if size is for data types such as `String`, `Array` etc and not for `Integer`/`Float`.

If I try using it with `Float`, I get the following message:
```
undefined method `size' for 12345678901234.0:Float
```

Using it on `Integer`s does not raise any errors, but the validation tests are failing.

PR #11, opened 2016-02-04
Invoice

ehannes:
I believe this branch is ready for code review. The only thing I know is missing is the repository spec and I'll notify when it's ready.


lib/fortnox/api/models/edi_information.rb
```diff
@@ -0,0 +1,28 @@
+require "fortnox/api/models/base"
+
+module Fortnox
+  module API
+    module Model
+      class EDIInformation < Base
```
d-Pixie: I prefer fully qualified paths for these, `Fortnox ::API::Model::Base`. Partially because when you have several Base classes in your tree it's not always obvious, without mental lookup, to see which one is used. It's easy to assume things ;) The other reason has to do with refactoring. If I move this class into another module later it might inherit from another Base and you can get weird errors etc.

ehannes:
Spec for InvoiceRepository added. Just copied the one from Customer. I think we can improve the repository specs further on :)

d-Pixie:
Yeah, and if we get a lot of duplication in the specs we should probably just create a spec for the base class instead and not test every use of it (which is basically what we are doing if all the specs turn out the same). Or at least move things up into shared example groups.

ehannes:
I think a shared context and shared example groups will do the trick.

lib/fortnox/api/validators/edi_information.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Fortnox
+  module API
+    module Validator
+      class EDIInformation
+
+        extend Fortnox::API::Validator::Base
+
+        using_validations do
+          # No validations needed, in other words always valid.
```
d-Pixie: Isn't it better to implement this as a separate class then? Something that has the same API as a real validator instance but just statically returns `true` for the `validate()` call?

lib/fortnox/api/validators/row.rb
```diff
@@ -12,10 +12,10 @@ class Row
           validates_length_of :description, length: 0..50, if: :description?
 
           validates_inclusion_of :account_number, within: (0..9999), if: :account_number?
-          validates_inclusion_of :delivered_quantity, within: (0..99_999_999_999_999), if: :delivered_quantity?
-          validates_inclusion_of :discount, within: (0..999_999_999_999), if: :discount?
+          validates_inclusion_of :delivered_quantity, within: (0..9_999_999_999_999.0), if: :delivered_quantity?
+          validates_inclusion_of :discount, within: (0..99_999_999_999.0), if: :discount?
           validates_inclusion_of :house_work_hours_to_report, within: (0..99_999), if: :house_work_hours_to_report?
-          validates_inclusion_of :price, within: (0..999_999_999_999), if: :price?
+          validates_inclusion_of :price, within: (0..99_999_999_999.0), if: :price?
```
d-Pixie: Good catch :)

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    ENV['FORTNOX_API_BASE_URL'] = ''
+    ENV['FORTNOX_API_CLIENT_SECRET'] = ''
+    ENV['FORTNOX_API_ACCESS_TOKEN'] = ''
+  }
+
+  after{
+    ENV['FORTNOX_API_BASE_URL'] = nil
+    ENV['FORTNOX_API_CLIENT_SECRET'] = nil
+    ENV['FORTNOX_API_ACCESS_TOKEN'] = nil
+  }
```
d-Pixie: These should probably be moved into a helper as well ... We'll need them for every repository spec.

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  let( :repository ){ Fortnox::API::Repository::Invoice.new }
+
+  describe '#save' do
+    context 'unsaved, new Invoice' do
```
d-Pixie: This context description is wrong, you are testing a saved invoice (or faking a saved invoice).

d-Pixie: I'm guessing it's my error from the beginning though :)

ehannes: As I said, I just copied the spec from Customer :). I'll fix it.

spec/fortnox/api/validators/edi_information_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+require 'fortnox/api/validators/context'
+require 'fortnox/api/validators/edi_information'
+require 'fortnox/api/models/edi_information'
+
+describe Fortnox::API::Validator::EDIInformation do
+  let( :model_class ){ Fortnox::API::Model::EDIInformation }
+  let( :valid_model ){ model_class.new }
```
d-Pixie: `model_class` is premature optimisation at this point. You only use it once.

spec/fortnox/api/validators/email_information_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+require 'fortnox/api/validators/context'
+require 'fortnox/api/validators/email_information'
+require 'fortnox/api/models/email_information'
+
+describe Fortnox::API::Validator::EmailInformation do
+  let( :model_class ){ Fortnox::API::Model::EmailInformation }
+  let( :valid_model ){ model_class.new }
```
d-Pixie: Same as above.

spec/fortnox/api/validators/edi_information_spec.rb
ehannes: True!

spec/fortnox/api/validators/email_information_spec.rb
ehannes: This is actually used a lot. Check out `spec/fortnox/api/validators/context.rb`

spec/fortnox/api/validators/invoice_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'spec_helper'
+require 'fortnox/api/models/invoice'
+require 'fortnox/api/validators/context'
+require 'fortnox/api/validators/invoice'
+
+describe Fortnox::API::Validator::Invoice do
+  let( :model_class ){ Fortnox::API::Model::Invoice }
+  let( :valid_model ){ model_class.new( customer_number: 12345 ) }
```
d-Pixie: Here you could argue it's ok (though it's a constant and constant lookup is pretty optimal as is in Ruby). The other problem with it is that you are basically creating a second pointer to a class and classes (all constants really) should begin with a capital letter. So if you want to do it you should really call it `Model` instead :)

spec/fortnox/api/validators/email_information_spec.rb
d-Pixie: Even worse then. You have a hidden dependency on the test creating the `model_class` variable. If I don't the context will fail it's tests I assume?

ehannes: Good point! The test will fail. But I think I can fix this...

d-Pixie: You should use:
```ruby
include_context "shared stuff" do
  let(:shared_let) { {'in_a' => 'block'} }
end
```
To make the association explicit.

d-Pixie: https://www.relishapp.com/rspec/rspec-core/docs/example-groups/shared-context second example.

ehannes: Yep, I know! :)

spec/fortnox/api/repositories/invoice_spec.rb
ehannes: Solved in 2dbfbfc.

lib/fortnox/api/validators/edi_information.rb
ehannes: Good idea. Then we would need some kind of super class for there two to guarantee that the API will stay the same over time, right?

d-Pixie: Adding a spec to make sure that the two classes respond to the same messages is enough. So no shared superclass, think duck typing, not inheritance.

Just verifying that the parts of the API that we care about matches between the two implementations is the right way to go here. It's like contracts in Go or interfaces in Java, but without runtime environment enforcement.

lib/fortnox/api/invoice.rb
```diff
@@ -0,0 +1,31 @@
+require "fortnox/api/models/invoice"
+require "fortnox/api/repositories/invoice"
+require "fortnox/api/validators/invoice"
+
+module Fortnox
+  module API
+    class Invoice < Fortnox::API::Model::Invoice
+
+      def valid?
+        Fortnox::API::Validator::Invoice.validate( self )
```
d-Pixie: Why not just `true`?

lib/fortnox/api/models/row.rb
```diff
@@ -3,7 +3,7 @@
 module Fortnox
   module API
     module Model
-      class Row < Base
+      class Row < Fortnox::API::Model::Base
```
d-Pixie: :+1:

lib/fortnox/api/invoice.rb
ehannes: Just copied `fortnox/api/customer.rb`.

lib/fortnox/api/repositories/invoice.rb
```diff
@@ -0,0 +1,35 @@
...
+              total_vat: "TotalVAT",
+              vat_included: "VATIncluded"
+            }
+          )
+        end
+
+      private
+
+        def instansiate( hash )
+          Fortnox::API::Model::Invoice.new( hash )
```
d-Pixie: This has changed. The base class knows how to do this now as long as you define the proper constant. Have a look at `Customer` for the new syntax for configuration and instantiation...

lib/fortnox/api/invoice.rb
d-Pixie: Sorry, I was in EDIInformation land :/ Ofc the Invoice model should be validated properly ...

lib/fortnox/api/validators/base.rb
```diff
@@ -6,7 +6,7 @@ module Validator
       module Base
 
         def validate( instance )
-          raise ArgumentError, "No validator given for #{name}" unless @validator
+          raise_error_if_no_validator
```
d-Pixie: I like the extraction.

lib/fortnox/api/validators/constant.rb
```diff
@@ -0,0 +1,20 @@
+module Fortnox
+  module API
+    module Validator
+      module Constant
+
+        def validate( _instance )
```
d-Pixie: Just `_` works for me. I know either is ok according to different coding standards, but you really don't care what it is you get in here ...

lib/fortnox/api/validators/constant.rb
```diff
@@ -0,0 +1,20 @@
+module Fortnox
+  module API
+    module Validator
+      module Constant
```
d-Pixie: Good

lib/fortnox/api/validators/invoice.rb
```diff
@@ -0,0 +1,34 @@
...
+          validates_length_of :customer_name, length: 0..1024, if: :customer_name?
+          validates_length_of :external_invoice_reference1, length: 0..80, if: :external_invoice_reference1?
+          validates_length_of :external_invoice_reference2, length: 0..80, if: :external_invoice_reference2?
+          validates_length_of :our_reference, length: 0..50, if: :our_reference?
+          validates_length_of :phone1, length: 0..1024, if: :phone1?
+          validates_length_of :phone2, length: 0..1024, if: :phone2?
+          validates_length_of :remarks, length: 0..1024, if: :remarks?
+          validates_length_of :your_order_number, length: 0..30, if: :your_order_number?
+          validates_length_of :your_reference, length: 0..50, if: :your_reference?
+          validates_length_of :zip_code, length: 0..1024, if: :zip_code?
```
d-Pixie: Maybe we could get this into nice columns?

```ruby
validates_inclusion_of :administration_fee,  within: (0..99_999_999_999.0),       if: :administration_fee?
validates_inclusion_of :currency_rate,       within: (0..999_999_999_999_999.0),  if: :currency_rate?
# ...

validates_length_of    :address1,  length: 0..1024,  if: :address1?
validates_length_of    :address2,  length: 0..1024,  if: :address2?
# ...
```

Something like that, note the double space between the different statements on the same line giving visual separation.

spec/fortnox/api/validators/base_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
+require 'fortnox/api/validators/base'
+require 'fortnox/api/validators/validator_examples'
+
+describe Fortnox::API::Validator::Base do
+  subject{ Class.new{ include Fortnox::API::Validator::Base }.new }
+
+  it_behaves_like 'validators' do
+    before do
+      subject.using_validations do
```
d-Pixie: Is this needed? What happens if you just leave it out?

spec/fortnox/api/validators/edi_information_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+describe Fortnox::API::Validator::EDIInformation do
+  let( :model_class ){ Fortnox::API::Model::EDIInformation }
+
+  include_context 'validator context' do
+    let( :valid_model ){ model_class.new }
+  end
+
+  describe '.validate EDIInformation' do
+    context 'with required attributes' do
```
d-Pixie: It doesn't have any required attributes ...

spec/fortnox/api/validators/email_information_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  describe '.validate EmailInformation' do
+    context 'with required attributes' do
+      it{ is_expected.to be_valid( valid_model ) }
+    end
+
+    include_examples 'validates length of string', :email_address_to, 1024
+    include_examples 'validates length of string', :email_address_cc, 1024
+    include_examples 'validates length of string', :email_address_bcc, 1024
+    include_examples 'validates length of string', :email_subject, 100
+    include_examples 'validates length of string', :email_body, 20000
```
d-Pixie: This is really neat, nice work :)

lib/fortnox/api/invoice.rb
ehannes: :D

spec/fortnox/api/validators/base_spec.rb
ehannes: You mean the `subject.using_validations` part? Then the test fails with:
```
Failure/Error: raise ArgumentError, "No validator given" unless @validator
```
The `validators` shared example assumes that the validatos is initialized.

spec/fortnox/api/validators/email_information_spec.rb
ehannes: Thanks!

spec/fortnox/api/validators/base_spec.rb
d-Pixie: Ah ok, just curious :)

PR #12, opened 2016-02-12
Customer validator spec uses shared context

ehannes:
When introducing shared context to the validators/customer_spec,
I found that validators for currency and country_code was
not working. These are separate attributes and are already validated.
For instance, if assigning 'A' to currency, it is still nil and no
validations are added. We must either change the attribute classes
and add the error there, or skip error messages for these attributes.

Additionally, validators context is updated with a shared examples
for validating inclusion of strings and the former shared examples
'validate inclusion of' is now called 'validate inclusion of number'.

### DIFF
--- a/lib/fortnox/api/validators/customer.rb
+++ b/lib/fortnox/api/validators/customer.rb
@@ -13,9 +13,6 @@ module Fortnox
 
           validates_presence_of :name
 
-          validates_length_of :currency,      length: 3,  if: :currency?
-          validates_length_of :country_code,  length: 2,  if: :country_code?
-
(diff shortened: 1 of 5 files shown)


ehannes:
Use attributes instead of base class.

ehannes:
Waiting for @d-Pixie to fix validations for attributes :)

d-Pixie:
So, validations for attributes are now fixed and merged in development. You'll have to change the base of this PR tough, since it's against master, and update it with the changes since it was open. Then we can merge it :)

ehannes:
Closing since we need to change base of this PR.

PR #13, opened 2016-02-12
Order added to project

ehannes:
Order and Invoice has a lot in common. The shared parts are placed
in OrderBase which both classes inherits from.


ehannes:
Not ready for merge yet. Just want your opinion, especially on the `OrderBase` class. Maybe there is a better way to deal with the shared code?

d-Pixie:
I'd go with a straight Ruby module and mix it in to be honest. This isn't really a superclass/subclass relationship, order/offer/invoice are not logically more specialised versions of the same base type the way chess is a subtype of board games, they just share traits.

It's generally better to favour composition over inheritance when you can since it gives you a lot more flexibility with your object graph. View the shared properties as just that, common traits, and put them in a module that you include in the other models.

d-Pixie:
https://youtu.be/v-2yFMzxqwU

ehannes:
I'll start working with this old PR again to implement `Order`. I'll refactor the code and make `OrderBase` a module rather than a class and include that module in `Invoice` and `Order`.


@d-Pixie: If `OrderBase` is to be a module, it can no longer inherit from `Fortnox::API::Model::Base` and will therefore have no access to the `Fortnox::API::Model::Base.attribute` method which declares Virtus attributes. What can I do about that? The method looks like this:

```
def self.attribute( name, *args )
  define_method( "#{name}?" ) do
  !send( name ).nil?

  super                                                                     
end
```

ehannes:
And another question: what should we call the module?

ehannes:
I just rebased this branch from `development` since it was really old, and now I see that this PR is opened directly to `master`. I'll close this PR and open a new one to `development`.

PR #14, opened 2016-02-26
Flaky tests

d-Pixie:
So, here is the branch with the wacky tests. I have tried a lot of stuff relating to how the dummy classes are declared and used in the tests, but I can't seem to get it working right again and I don't see anything that looks off.

What I have concluded is that the tests are deterministic, the first in each of currency and country_code attributes spec works, the rest fails (it's like it's not calling the constructor at all for any but the first). For the new tests, the attribute validations, I have no idea. I think it's the same issue affecting both and the order the suit is run determines if the attribute validators pass or fail.

I get anywhere from 3 to 5 failed tests locally...

### DIFF
--- a/lib/fortnox/api/models/attributes/country_code.rb
+++ b/lib/fortnox/api/models/attributes/country_code.rb
@@ -6,15 +6,12 @@ module Fortnox
       module Attribute
         module CountryCode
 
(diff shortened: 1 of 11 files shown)


d-Pixie:
Ok, so this commit extracts the helper into the Helpers module and refactors the tests to use it. Should solve all the problems and looks reasonably nice compared to some of the alternatives :)

d-Pixie:
Well .... Honestly I can't see how to make Code Climate happy for the remaining 4 issues ... Not without contorting the code just to please Code Climate and actually reducing the code quality in the mean while :/

Any thoughts on it @ehannes ?

d-Pixie:
Also; Why arn't the new commits showing up under `commits` nor `changed files`? They are counted in the PR going by the `d-Pixie  wants to merge 10 commits into master from flaky_tests` at the top. But the commits only show 4 commits and the changed files are only for those commits ...

ehannes:
I'll have a look at CodeClimate :)

ehannes:
How about this to get rid of code duplications in attribute validators specs?

`spec/fortnox/api/validators/attributes/context.rb`:
```
shared_context 'attributes context' do |attribute_class, validator_class|
  using_test_classes do
    class Model < Fortnox::API::Model::Base
      include attribute_class
    end

    class Validator
      extend Fortnox::API::Validator::Base
      include validator_class
    end
  end
end
```

`spec/fortnox/api/validators/attributes/currency_spec.rb`:
```
describe Fortnox::API::Validator::Attribute::Currency do
  include_context 'attribute context', 
    Fortnox::API::Model::Attribute::Currency,
    Fortnox::API::Validator::Attribute::Currency

    # Tests...
end
```

ehannes:
Or even better if you can get this to work:
```
describe Fortnox::API::Validator::Attribute::Currency do
  include_context 'attribute context', Currency

    # Tests...
end
```

ehannes:
The code duplication in `lib/fortnox/api/models/attributes/country_code.rb` is tricky. Have no idea of how to get rid of it.

d-Pixie:
I'm not a fan of hiding stuff, unless it's details, just for the sake of it...

I read it wrong (was on my phone) btw, your second proposal above, might work. What I tried wasn't the same :)

Feel free to try it out and let's see how the tests look after. The repeated block for the module chain I don't think we can get away from. ..

Maybe time to revisit that idea of excluding CC form the build :) We can always run it separately t check the tests....

d-Pixie:
Also I'll have to close this PR and open a new one against development...

PR #16, opened 2016-02-29
Flaky tests

d-Pixie:
_Reopening of PR https://github.com/my-codeworks/fortnox-api/pull/14 against the correct branch. This was necessary since we changed the branching scheme after the PR was originally opened._

So, here is the branch with the wacky tests. I have tried a lot of stuff relating to how the dummy classes are declared and used in the tests, but I can't seem to get it working right again and I don't see anything that looks off.

What I have concluded is that the tests are deterministic, the first in each of currency and country_code attributes spec works, the rest fails (it's like it's not calling the constructor at all for any but the first). For the new tests, the attribute validations, I have no idea. I think it's the same issue affecting both and the order the suit is run determines if the attribute validators pass or fail.

I get anywhere from 3 to 5 failed tests locally...

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -6,11 +6,12 @@ engines:
     enabled: true
     config:
       languages:
-      - ruby
-      exclude_paths:
-      # Exclude spec/ from duplication engine since too dry specs
-      # reduces readability.
-      - spec/
+        ruby:
+          mass_threshold: 20
+          exclude_paths:
+            # Exclude spec/ from duplication engine since too dry specs
+            # reduces readability.
+            - spec/
   eslint:
     enabled: true
   fixme:
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,6 +7,10 @@ engines:
     config:
       languages:
       - ruby
+      exclude_paths:
+      # Exclude spec/ from duplication engine since too dry specs
+      # reduces readability.
+      - spec/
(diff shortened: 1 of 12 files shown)


d-Pixie:
This reopen makes it possible for me to merge this to development even with the outstanding CC issues :) We won't be able to merge development down to master with them still there though. Maybe we should relax that? Not that I want to allow crap through, but CC isn't perfect and at least two of the issues on this branch are hard to fix in a sane way and any solution I come up with is going to make the code worse just to make CC pass, and that's just wrong.

ehannes:
Can you ignore issues in CC?

ehannes:
We could [exclude `specs` directory](https://docs.codeclimate.com/docs/excluding-files-and-folders) for the `duplication` engine, or even narrow it down to maybe just the `/spec/fortnox/api/validators/attributes` folder... Something like this:
`.codeclimate.yml`:
```
---
engines:
  bundler-audit:
    enabled: true
  duplication:
    enabled: true
    config:
      languages:
     - ruby
      exclude_paths:
    - spec/
# Etc...
```

And the other issues found in `lib/fortnox/api/models/attributes/currency.rb` and the other attribute class, we can raise the mass level from 18 to maybe 20.

ehannes:
CodeClimate is a bit slow... It hasn't run since the last commit... Should still be 2 unresolved issues.

PR #17, opened 2016-03-01
Test classes context added for attribute specs

ehannes:
Removed code duplication from attributes specs

### DIFF
--- a/spec/fortnox/api/models/attributes/test_classes_context.rb
+++ b/spec/fortnox/api/models/attributes/test_classes_context.rb
@@ -1,0 +1,9 @@
+shared_context 'test classes' do
+  using_test_class do
+    class Model
+      include Virtus.model
+    end
+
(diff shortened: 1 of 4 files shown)


spec/fortnox/api/models/attributes/test_classes_context.rb
```diff
@@ -0,0 +1,9 @@
+shared_context 'test classes' do
```
d-Pixie: I like the setup, but I don't like the name :) It doesn't say anything about what that row does. Imagine that you just came on the project and looked at the `include_context 'test classes'` statement in the specs, what would it tell you?

Something like `shared_context 'create dummy Model that includes the described attribute' do` would probably mean you don't have to go look at the implementation ...

ehannes: I completely agree!

PR #21, opened 2016-03-02
Customer validator spec uses shared context

ehannes:
Reopening of #12 but now against `development` branch instead of `master`.

Original message:
When introducing shared context to the validators/customer_spec,
I found that validators for currency and country_code was
not working. These are separate attributes and are already validated.
For instance, if assigning 'A' to currency, it is still nil and no
validations are added. We must either change the attribute classes
and add the error there, or skip error messages for these attributes.

Additionally, validators context is updated with a shared examples
for validating inclusion of strings and the former shared examples
'validate inclusion of' is now called 'validate inclusion of number'.

### DIFF
--- a/lib/fortnox/api/validators/customer.rb
+++ b/lib/fortnox/api/validators/customer.rb
@@ -8,12 +8,13 @@ module Fortnox
         using_validations do
 
           VAT_TYPES = ['SEVAT', 'SEREVERSEDVAT', 'EUREVERSEDVAT', 'EUVAT', 'EXPORT']
+          TYPES = ['PRIVATE', 'COMPANY']
 
           validates_presence_of :name
 
-          validates_inclusion_of :sales_account,  within: (0..9999),               if: :sales_account?
-          validates_inclusion_of :type,           within: ['PRIVATE', 'COMPANY'],  if: :type?
-          validates_inclusion_of :vat_type,       within: VAT_TYPES,               if: :vat_type?
+          validates_inclusion_of :sales_account,  within: (0..9999),  if: :sales_account?
+          validates_inclusion_of :type,           within: TYPES,      if: :type?
+          validates_inclusion_of :vat_type,       within: VAT_TYPES,  if: :vat_type?
 
         end
 
--- a/lib/fortnox/api/validators/customer.rb
+++ b/lib/fortnox/api/validators/customer.rb
@@ -11,9 +11,6 @@ module Fortnox
 
           validates_presence_of :name
 
-          validates_length_of :currency,      length: 3,  if: :currency?
-          validates_length_of :country_code,  length: 2,  if: :country_code?
-
           validates_inclusion_of :sales_account,  within: (0..9999),               if: :sales_account?
           validates_inclusion_of :type,           within: ['PRIVATE', 'COMPANY'],  if: :type?
           validates_inclusion_of :vat_type,       within: VAT_TYPES,               if: :vat_type?
--- a/lib/fortnox/api/validators/customer.rb
+++ b/lib/fortnox/api/validators/customer.rb
@@ -1,9 +1,14 @@
 require "fortnox/api/validators/base"
+require "fortnox/api/validators/attributes/country_code"
+require "fortnox/api/validators/attributes/currency"
 
 module Fortnox
   module API
     module Validator
       class Customer < Fortnox::API::Validator::Base
+
+        include Fortnox::API::Validator::Attribute::CountryCode
+        include Fortnox::API::Validator::Attribute::Currency
 
         using_validations do
 
--- a/spec/fortnox/api/validators/context.rb
+++ b/spec/fortnox/api/validators/context.rb
@@ -13,7 +13,7 @@ shared_context 'validator context' do
   end
   shared_examples_for 'invalid' do |attribute, values, error_type|
     values.each do |value|
-      context "when #{attribute} set to #{value}" do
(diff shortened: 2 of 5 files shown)


ehannes:
I'm not sure what we said about attribute validations. How are we going to use them? This PR removes the following from `Fortnox::API::Validator::Customer`:
```
validates_length_of :currency,      length: 3,  if: :currency?
validates_length_of :country_code,  length: 2,  if: :country_code?
```
Should we include the `currency` and `country_code` validators in `Fortnox::API::Validator::Customer`?

d-Pixie:
Yes, the plan was to mix the attribute validators into the main validator the same way we mix in the attributes into models.

ehannes:
Should we merge now or should I include the validators before merging?

d-Pixie:
Refactor to include the new validators before merge please. Otherwise we'll just get that change in an unrelated PR later ...

ehannes:
Issue #27 makes tests failing...

lib/fortnox/api/validators/customer.rb
```diff
@@ -1,22 +1,25 @@
...
 
           validates_presence_of :name
 
-          validates_length_of :currency,      length: 3,  if: :currency?
-          validates_length_of :country_code,  length: 2,  if: :country_code?
-
-          validates_inclusion_of :sales_account,  within: (0..9999),               if: :sales_account?
-          validates_inclusion_of :type,           within: ['PRIVATE', 'COMPANY'],  if: :type?
-          validates_inclusion_of :vat_type,       within: VAT_TYPES,               if: :vat_type?
+          validates_inclusion_of :sales_account,  within: (0..9999),  if: :sales_account?
```
d-Pixie: We should really extract account into it's own attribute class. Only problem is how to name it :) We might get the same problem with country code and currency as well, if they aren't always called exactly that. I'll have a look at it, just ignore it for this PR and we'll refactor when I have a solution for it.

spec/fortnox/api/validators/context.rb
```diff
@@ -13,7 +13,7 @@
   end
   shared_examples_for 'invalid' do |attribute, values, error_type|
     values.each do |value|
-      context "when #{attribute} set to #{value}" do
+      context "when :#{attribute} set to #{value}" do
```
d-Pixie: This is fine, if we are sure they will always be symbols. If you wanted something more generic you could do `"#{ attribute.inspect }"` instead so that it prints the value with the right representation regardless of class.

ehannes: This is only how to display the name of the attribute, not what type the attribute holds.

spec/fortnox/api/validators/customer_spec.rb
```diff
@@ -1,33 +1,31 @@
...
-      let( :customer ){ Fortnox::API::Model::Customer.new( name: 'Test' ) }
-
-      it 'is valid' do
-        expect( subject.validate( customer )).to eql( true )
-      end
-    end
-
-    context 'Customer with invalid sales_account' do
-      let( :customer ){ Fortnox::API::Model::Customer.new( name: 'Test', sales_account: 99999 ) }
+  let( :model_class ){ Fortnox::API::Model::Customer }
```
d-Pixie: I don't like this tbh. It's obfuscating and only serves the purpose of saving you a search and replace with multiple hits instead of one hit if we ever change the class name. A better option, if you want something shorter, is to do either:
```ruby
using_test_class do
  Customer = Fortnox::API::Model::Customer
end
```
or
```
include Fortnox::API::Model
```
where the last one imports all the constants from `Fortnox::API::Model` into the local scope, so you get access to `Model` in the local scope instead of going through the `Fortnox::API::Model` chain explicitly each time. You might have to wrap this in `using_test_class` as well to ensure no constant overload warnings/errors ...

spec/fortnox/api/validators/invoice_spec.rb
```diff
@@ -34,9 +28,9 @@
...
     include_examples 'validates length of string', :zip_code, 1024
 
-    include_examples 'validates inclusion of', :administration_fee, 0, 99_999_999_999.0
-    include_examples 'validates inclusion of', :currency_rate, 0, 999_999_999_999_999.0
-    include_examples 'validates inclusion of', :currency_unit, 0, 999_999_999_999_999.0
-    include_examples 'validates inclusion of', :freight, 0, 99_999_999_999.0
+    include_examples 'validates inclusion of number', :administration_fee, 0, 99_999_999_999.0
+    include_examples 'validates inclusion of number', :currency_rate, 0, 999_999_999_999_999.0
+    include_examples 'validates inclusion of number', :currency_unit, 0, 999_999_999_999_999.0
+    include_examples 'validates inclusion of number', :freight, 0, 99_999_999_999.0
```
d-Pixie: Really need to make a `validates coverage of number` version of this ... I'll put it on my todo.

spec/fortnox/api/validators/context.rb
d-Pixie: Yes, I know. You are printing it as: `"This is an :attribute"`, which is fine as long as we know we will always get a symbol. If you do `"This is an #{ attribute.inspect }"` it would still be `"This is an :attribute"` if the `attribute` variable contained a symbol but would be `"This is an \"attribute\""` if the attribute var was a string instead etc ...

ehannes: I get it! Great point! I'll update it with `#{ attribute.inspect }`.

Issue #22, opened 2016-03-02
Error when calling Fortnox::API::Customer.all


ehannes:
To reproduce the problem:

```
bundle exec irb
```

Then run the following command in irb (replace [path_to_gem]):
```
irb(main):001:0> $: << "[path_to_gem]/lib"; require 'fortnox/api'
=> true
irb(main):002:0> Fortnox::API::Customer.all
ArgumentError: bad argument (expected URI object or URI string)
        from ~/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/httparty-0.13.7/lib/httparty/request.rb:55:in `path='
        from ~/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/httparty-0.13.7/lib/httparty/request.rb:43:in `initialize'
        from ~/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/httparty-0.13.7/lib/httparty.rb:545:in `new'
        from ~/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/httparty-0.13.7/lib/httparty.rb:545:in `perform_request'
        from ~/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/httparty-0.13.7/lib/httparty.rb:476:in `get'
        from [path_to_gem]/lib/fortnox/api/base.rb:42:in `block (3 levels) in <class:Base>'
        from [path_to_gem]/lib/fortnox/api/request_handling.rb:29:in `call'
        from [path_to_gem]/lib/fortnox/api/request_handling.rb:29:in `execute'
        from [path_to_gem]/lib/fortnox/api/base.rb:41:in `block (2 levels) in <class:Base>'
        from [path_to_gem]/lib/fortnox/api/repositories/base/loaders.rb:9:in `all'
        from [path_to_gem]/lib/fortnox/api/customer.rb:18:in `all'
        from (irb):2
        from ~/.rbenv/versions/2.2.3/bin/irb:11:in `<main>'
```

ehannes:
Calling `Fortnox::API::Customer.find(1)` gives another error message, but they could be linked:
```
irb(main):003:0> Fortnox::API::Customer.find(1)
NoMethodError: undefined method `+' for nil:NilClass
        from [path_to_gem]/lib/fortnox/api/repositories/base/loaders.rb:35:in `find_one_by'
[...]

PR #23, opened 2016-03-07
Fix 22

ehannes:
Not ready for merge yet. Just want you to see what I am working with.


ehannes:
Do you have some idea for better names for methods in `Fortnox::API::InputValidation`?

d-Pixie:
I think that renaming the module to `Environment` or some such might be a good idea and then rename the methods to `get_base_url` etc. Or do you have any preference?

ehannes:
Methods are named `get_base_url` etc. in e7fbd16 and the module is called `EnvironmentValidation`. Maybe `Environment` is enough?

d-Pixie:
Always refrence the issue you are solving in the PR like this (just paste the full URL to the issue and Github does the rest) https://github.com/my-codeworks/fortnox-api/issues/22 ... That way they become linked and I can see in the issue that there is a PR for it and go directly to the issue from the PR as well :)

d-Pixie:
I have no strong preference for `Environment` vs `EnvironmentValidation `. The former is shorter and pretty clear. The latter is clearer but I'm not sure it adds any clarity since it's still not obvious that the methods come from this module.

ehannes:
#680f412 includes fix for #29

ehannes:
I think it's time for code review this branch now since all specs passes at last! Would you mind taking a look @d-Pixie? :)

lib/fortnox/api/models/base.rb
```diff
@@ -36,7 +36,9 @@ def update( hash )
 
           return self if new_attributes == old_attributes
 
-          self.class.new( new_attributes.select{ |_,v| !v.nil? })
+          new_hash = new_attributes.select{ |_,v| !v.nil? }
```
d-Pixie: Better as: `new_attributes.delete_if{ |_, value| value.nil? }`

lib/fortnox/api/repositories/base/loaders.rb
```diff
@@ -27,19 +27,19 @@ def find( id_or_hash )
...
 
-        catch ArgumentError
+        rescue ArgumentError
           raise ArgumentError, "find only accepts a number or hash as argument"
         end
 
         def find_one_by( id )
-          response_hash = get( @uri + id )
-          entity_hash = response_hash[ @json_entity_wrapper ]
+          response_hash = get( @options.uri + id.to_s )
```
d-Pixie: Use interpolation instead of manual casting and concatentation: `response_hash = get( "#{ @options.uri }#{ id }" )`

lib/fortnox/api/repositories/base/savers.rb
```diff
@@ -9,19 +9,30 @@ def save( entity )
           hash = entity_to_hash( entity )
           json = hash.to_json
 
-          response = entity.new? ? save_new( json ) : update_existing( json )
+          entity.new? ? save_new( json ) : update_existing( hash )
```
d-Pixie: What about:
```
return save_new( json ) if entity.new?
update_existing( json )
```
You are converting to json even if you don't need it now and you are creating two sub methods with different levels of abstraction (one on the level of a hash and the other on the level of json).

lib/fortnox/api/repositories/base/savers.rb
```diff
@@ -9,19 +9,30 @@ def save( entity )
...
         end
 
       private
 
         def save_new( json )
-          post( @uri, { body: json } )
+          post( @options.uri, { body: json } )
```
d-Pixie: This then needs: `json = hash.to_json` ofc.

spec/fortnox/api/models/base_spec.rb
```diff
@@ -1,52 +1,75 @@
...
-class TestEntity < Fortnox::API::Model::Base
-  attribute :private, String, writer: :private
-  attribute :string, String
-  attribute :number, Integer, default: 42
-end
-
 describe Fortnox::API::Model::Base do
-  let(:original){ TestEntity.new( string: 'Test' ) }
+  using_test_classes do
+    class TestEntity < Fortnox::API::Model::Base
```
d-Pixie: I really do not like prefixes like this :) It's obviously a test entity, it's in a test (and we don't have any other model named `Entity` in our gem).

d-Pixie:
Minor style stuff, only one real objection (in savers), the rest is just opinions :) Fix that and it will look good to me.

lib/fortnox/api/repositories/base/savers.rb
ehannes: You mean like this:

```
hash = entity_to_hash( entity )

return save_new( hash ) if entity.new?
update_existing( hash )
```

And then `save_new` converts to json as in your next comment?

d-Pixie: Eh, yeah. Just like that :) Sorry about the hash/json mixup ...

Issue #27, opened 2016-03-08
Attribute validators not working


ehannes:
Attribute validators are not working as intended. Validations included form an attribute module is overwritten by the validations added in the validator including the module.

For instance, adding the following to `validators/attributes/country_code_spec.rb` will make the test fail:

```
class TestValidator
  extend Fortnox::API::Validator::Base
  include Fortnox::API::Validator::Attribute::CountryCode

  using_validations do
  end
end
```

ehannes:
I'm affraid #28 did not solve our problem entirely.

When trying to get attribute validation work together with another resource, `Customer` in my case, I get strange errors.
```
  3) Fortnox::API::Validator::Customer.validate Customer with country_code set to a nonsense string should include error for country_code
     Failure/Error: it{ is_expected.to include_error_for( updated_model, attribute, error_type ) }
[...]

PR #28, opened 2016-03-08
Adds support for multiple using_validations block

d-Pixie:
Changes the validation mechanism to allow for many validator blocks begin called
during validator construction and serving the aggregate results of all of them
as the result of the validation.

Should fix https://github.com/my-codeworks/fortnox-api/issues/27 ...


ehannes:
Looking great!

ehannes:
Works like a charm! I'll merge if Travis is satisfied.

PR #33, opened 2016-03-11
Wrapper classes removed. Fixes #31

### LINKED ISSUE #31: Remove wrapper class for Customer
`Fortnox::API::Customer` should be removed. Instead, use `Fortnox::API::Model::Customer`, `Fortnox::API::Validator::Customer` and `Fortnox::API::Repository::Customer`.

ehannes:
Fixes #31 

With the new architecture for the gem, wrapper classes are not
needed anymore.

To load all dependencies, api requires three new classes:
Fortnox::API::Models, Fortnox::API::Repositories,
Fortnox::API::Validators. These classes then loads required model,
repository and validator classes.


lib/fortnox/api/models.rb
```diff
@@ -0,0 +1,8 @@
+require "fortnox/api/models/customer"
+
+module Fortnox
```
d-Pixie: This is not needed.

lib/fortnox/api/repositories.rb
```diff
@@ -0,0 +1,8 @@
+require "fortnox/api/repositories/customer"
+
+module Fortnox
```
d-Pixie: Same here.

lib/fortnox/api/validators.rb
```diff
@@ -0,0 +1,8 @@
+require "fortnox/api/validators/customer"
+
+module Fortnox
```
d-Pixie: And here ... :)

d-Pixie:
If we had used the `class Fortnox::API::Model::Customer ... end` syntax to declare the models/validators/repositories we would have had to initialize the parent modules in the path first. But since we use explicit module definitions in each of the classes:
```ruby
module Fortnox
  module API
    module Model
      class Customer
        ...
      end
    end
  end
end
```
Each class declaration essentially redeclares it's module chain this way, so you never have to worry about when the initilisation is run...

ehannes:
I'm not sure I'm following...

If `lib/fortnox/api/models.rb` does not need this code:
```require "fortnox/api/models/customer"
+
+module Fortnox
```

d-Pixie:
No, the commented line is the first line in this block:
```ruby
module Fortnox
  module API
    module Model
    end
  end
end
```
That's what's not needed, since the required file already declares this anyway.

ehannes:
Ok. So should I remove the whole block? And just keep the first line in `lib/fortnox/api/models.rb`, like this:

```
require "fortnox/api/models/customer"
```

d-Pixie:
Yep :)

PR #48, opened 2016-03-16
Invoice repo spec

ehannes:
An attempt to refactor the customer repository spec.

The customer repository has pretty good test coverage. The other repositories does not. This is an attempt to refactor the customer repository spec and extract tests to shared examples so all repositories can include these tests.

The first repository to include these shared examples, besides customer repository, is invoice repository.


ehannes:
This is a first attempt to extract repository tests into shared examples. @d-Pixie, please have a look at `spec/fortnox/api/repositories/customer` and have share your thoughts on that. There are many details that can be improved, for instance the shared examples takes very many parameters...

d-Pixie:
Ok, firstly: great initiative :) To solve some of the problems then: `Fortnox::API::Repository::Customer` has a `MODEL` constant that you can use to get the correct class for the repositories model so you don't have to send that in. You might also consider accessing the `@options` for the repository and use things like `@options.json_collection_wrapper.downcase` as the VCR directory, or derive it from the name in the `MODEL`constant.

ehannes:
Do I have access to `@options`? If I do
```
it{ puts subject.options }
```
I get:
```
  1) Fortnox::API::Repository::Customer 
     Failure/Error: it{ puts subject.options }
     
     NoMethodError:
       undefined method `options' for #<Fortnox::API::Repository::Customer:0x007f560b326d98>
     # ./spec/fortnox/api/repositories/customer_spec.rb:24:in `block (2 levels) in <top (required)>'
```

d-Pixie:
No, but you can get access by adding `attr_accessor :options` to the repository base class ...

ehannes:
Or should I include `attr_reader :options` instead?

d-Pixie:
Yes, that is probably a better idea :D

ehannes:
There are some old tests situated in `spec/fortnox/api/repository/examples.rb`. I have created that shared examples file, but if I'm not mistaken, they are originally written by you @d-Pixie. Now that we are using VCR, do we need these tests any more or can I delete them? Or should we rewrite them in some way?

```
shared_examples_for 'repositories' do |model_class|
  describe '#save' do
    context "new #{model_class}" do
      let( :repository ){ described_class.new }
      let( :model ){ model_class.new( unsaved: false ) }
 
      before do
        # Should not make an API request in test!
        expect( repository ).not_to receive( :save_new )
        expect( repository ).not_to receive( :update_existing )
      end
 
      specify{ expect( repository.save( model )).to eql( true ) }
    end
  end
end
```

d-Pixie:
I will assume that your new cases cover whatever was in the `spec/fortnox/api/repository/examples.rb` file? And it wasn't me: https://github.com/my-codeworks/fortnox-api/blame/development/spec/fortnox/api/repositories/examples.rb :)

spec/fortnox/api/repositories/examples/save_examples.rb
```diff
@@ -0,0 +1,41 @@
+shared_examples_for '.save' do |attribute, fortnox_attribute|
```
d-Pixie: The `fortnox_attribute` is the same as `described_class.options.json_entity_wrapper`no?

d-Pixie: Or `subject.options.json_entity_wrapper` ...

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -1,10 +1,25 @@
...
+  it_behaves_like 'repositories'
+
+  include_examples '.all'
+
+  include_examples '.find', :document_number
+
+  include_examples '.save', :comments, 'Comments' do
+    let( :attribute_value ){ 'An invoice comment' }
+    let( :updated_attribute_value ){ 'An updated invoice comment' }
+    let( :valid_model ) do
```
d-Pixie: Can't this be done inside the example group? It will be the same for all no? Except that you will have to make it:
```
attributes = {}
attributes[ attribute ] = attribute_value
attributes[ :customer_number ] = 1
described_class::MODEL.new( attributes )

File: spec/fortnox/api/repositories/examples/save_examples.rb (hunk 1 above)
Comment (d-Pixie) [inline spec/fortnox/api/repositories/examples/save_examples.rb]: No, I was wrong :)

Comment (d-Pixie) [inline spec/fortnox/api/repositories/examples/save_examples.rb]: I don't think that `attribute` and `fortnox_attribute` are good names. I'm guessing that `attribute` should be `attribute_name` or some such? And for `fortnox_attribute` I'd go with `attribute_json_name` ... And that leads me to want them both at the same level of specificity so then I'd change `attribute` to `attribute_hash_name` or `attribute_hash_key_name` instead.

The main point of these shared examples is that since they are out of context when you read them (the incoming actual data is in another file) you need to be even more careful in how you name things and generally opt for more specific names when you can.

Comment (d-Pixie) [inline spec/fortnox/api/repositories/examples/save_examples.rb]: As a further comment on this I think you should only take the `attribute_hash_key_name` as an argument since the repository in question already has functions that map from a JSON key name to a hash key name, use them instead :)

File: spec/fortnox/api/repositories/invoice_spec.rb [hunk 3]
```diff
@@ -1,10 +1,25 @@
...
   include_context 'repository context'
 
-  it_behaves_like 'repositories', Fortnox::API::Model::Invoice
+  it_behaves_like 'repositories'
+
+  include_examples '.all'
+
+  include_examples '.find', :document_number
+
+  include_examples '.save', :comments, 'Comments' do
```
d-Pixie: With all of the related comments implemented this would become:
```
include_examples '.save', :comments do
  let( :value ){ 'An invoice comment' }
  let( :updated_value ){ 'An updated invoice comment' }
end
```

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -1,10 +1,25 @@
...
+    let( :attribute_value ){ 'An invoice comment' }
+    let( :updated_attribute_value ){ 'An updated invoice comment' }
```
d-Pixie: Skip the `attribute_` part of the names.

spec/fortnox/api/repositories/examples/find_examples.rb
```diff
@@ -0,0 +1,25 @@
+shared_examples_for '.find' do |id_attribute|
```
d-Pixie: The id attribute is also available on the repository, we should not ask for it again.

d-Pixie:
I had some comments for refactoring this further using more of the information that is already declared in the repository.

spec/fortnox/api/repositories/invoice_spec.rb
ehannes: As I said, I created a shared example, but you created the test in the first place: https://github.com/my-codeworks/fortnox-api/commit/53bb403869cf8cb41c18256ea71f5816f0a7c611

So, as it was you who initially wrote the test, I was hoping that you could explain what the test actually test.
```
before do
  # Should not make an API request in test!
  expect( repository ).not_to receive( :save_new )
  expect( repository ).not_to receive( :update_existing )
end

specify{ expect( repository.save( model )).to eql( true ) }
```
I assume that `# Should not make an API request in test!` was before we used VCR? If I read the test correctly, we just check that when calling `repository.save( model )` it returns `true`, right? I don't think we need to check that anymore since we use VCR. If `save` returns `false`, something is wrong with the request and VCR will nofity.

ehannes: About `let( :valid_model ) do`, I can do something like:

```
let( :valid_hash ){ comments: attribute_value, customer_number: 1 }
```
```
let( valid_hash){ name: attribute_value }
```

spec/fortnox/api/repositories/examples/save_examples.rb
ehannes: Great idéas!

spec/fortnox/api/repositories/examples/find_examples.rb
ehannes: If not, we need to convert the attribute name from JSON format (as stated in the repository) to a symbol. I believe this functionality is built into the repository as a private method? Should we make that method public so we can use it here in the test? Or is it simple to write?

d-Pixie:
I think what's tripping you up is that I have a typo in the context declaration, this: `context 'unsaved, new customer' do` should be: `context 'saved, new customer' do`. The customer has `unsaved: false` which == `saved: true`. The test just makes ure that if a customer is already saved it is not saved again. No call to either save or update is made.

ehannes:
That ^ made things a lot clearer! Then I understand what the test does! I move this test into the `.save` shared examples.

ehannes:
With your help @d-Pixie, I think the repository specs are very clean and nice to read! I created a shared context and included `JSONHelper` as a test class. Take a look at https://github.com/my-codeworks/fortnox-api/pull/48/commits/e455d4bb39bf8bafb1092100271b1673de2a8e2f if you want. If you do not have any more comments, I think this PR is ready for merge! :dancers:

spec/fortnox/api/repositories/examples/find_examples.rb
```diff
@@ -6,11 +9,15 @@
...
   describe '.find' do
+    include_context 'JSONHelper'
+
     specify 'returns correct class' do
       expect( find_id_1.class ).to be described_class::MODEL
     end
 
     specify 'returns correct Customer' do
+      id_attribute_json = subject.options.unique_id
+      id_attribute = JSONHelper.new.convert_from_json(id_attribute_json)
```
d-Pixie: Add `module_function :method` after the declaration of the methods you want to use from the module instead. That will allow you to do `Fortnox::API::Repository::JSONConvertion.convert_from_json( id_attribute_json )` directly without having to wrap it in a new class and instantiating it.

See http://apidock.com/ruby/Module/module_function for docs :)

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -1,10 +1,20 @@
...
 
-  it_behaves_like 'repositories', Fortnox::API::Model::Invoice
+  include_examples '.all'
+
+  include_examples '.find'
+
+  include_examples '.save', :comments do
+    let( :value ){ 'An invoice comment' }
+    let( :updated_value ){ 'An updated invoice comment' }
+    let( :model_hash ){ { comments: value, customer_number: 1 } }
```
d-Pixie: Why are we still creating this outside the example group? Can't you just include this in the example group as well? If you need to send the id in (which I guess you don't right now) you can always add it as another prop you send in.

I'm also thinking that the value isn't that important? The specific value I mean. As long as the example group throws an exception if the type of the value isn't a string (or doesn't respond to `.to_s`) we should be safe. Most properties on models are strings so finding a property for the test for any model that can take a random sting shouldn't be a problem. That way you can just use `"A value"` and `"A changed value"` statically in the example group and this would all just become:

```
include_examples '.save', :comments
```
Like the rest of the include_example lines ...

Arguments against this are welcome however. I'm not sure it's a good idea, but it would make the tests consistent and nice looking :)

spec/fortnox/api/repositories/examples/find_examples.rb
ehannes: Great idea, but I could not get it to work. `convert_key_from_json` uses `@options`, which is set when the repository is initialized...

spec/fortnox/api/repositories/invoice_spec.rb
ehannes: If we can assume that all models have a text field that can include whatever text we provide, then I guess we can safely remove `:value` and `:updated_value`.

I think it is harder to remove `model_hash`. That hash will look differently depending on what attributes are required for the current model. For instance, `Customer` requires `:name` and `Invoice` requires `customer_number`. But if you have any solution to this, please let me know and I'll implement it :)

ehannes:
One unsolved comment from above disappeared since I renamed the examples:

You suggested that I should: _Add module_function :method after the declaration of the methods you want to use from the module instead._

I responded with: _Great idea, but I could not get it to work. `convert_key_from_json` uses `@options`, which is set when the repository is initialized..._

ehannes:
And now you can take a look at https://github.com/my-codeworks/fortnox-api/pull/48/files/e455d4bb39bf8bafb1092100271b1673de2a8e2f#r56538102 and see what you think.

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -1,10 +1,16 @@
...
 describe Fortnox::API::Repository::Invoice do
-  include_context 'repository context'
+  include_context 'environment'
 
-  it_behaves_like 'repositories', Fortnox::API::Model::Invoice
+  include_examples '.all'
+
+  include_examples '.find'
+
+  include_examples '.save', :comments, { customer_number: 1 }
```
d-Pixie: That is a good idea, to take a hash as the last argument, I like it :)

spec/fortnox/api/repositories/customer_spec.rb
```diff
@@ -1,85 +1,16 @@
...
-        let( :model ){ find_customer_1.update( name: updated_attribute ) }
-
-        let( :send_request ) do
-          VCR.use_cassette( 'customers/save_old' ){ subject.save( model ) }
-        end
-        let( :response ){ send_request }
-      end
-    end
-  end
+  include_examples '.save', :name, {}
```
d-Pixie: Don't do this, see comment on the save example group.

spec/fortnox/api/repositories/examples/save.rb
```diff
@@ -0,0 +1,61 @@
+# Assumes that attribute_hash_name holds a string without restrictions.
+shared_examples_for '.save' do |attribute_hash_name, required_attributes|
```
d-Pixie: Do this instead: `shared_examples_for '.save' do |attribute_hash_name, required_attributes = {}|`

ehannes: Nice one!

d-Pixie:
Ah, as I said before I thought the JSON helpers where standalone ... What you could do is refactor the code in the helpers so that the parent method, `convert_keys_from_json`, sends the part of the options that you need down to `convert_key_from_json`. That way you can use the singular method as standalone tool. We might actually continue that refactoring later and make it so that all the state needed by the JSON helpers is sent in. Makes sense since we already have utility of the methods outside of the Repository context. OR ... we just make them public on the repository class, that is fine as well ...

d-Pixie:
New issue, https://github.com/my-codeworks/fortnox-api/issues/52, to track the refactoring. I see you added default params to the block. Isn't this ready for merge then?

ehannes:
The default block is added. If we can leave `JSONHelper` in the repository specs for now, then we can merge this :)

d-Pixie:
Yes I think we should since there is a standalone issue to handle that later. I'll add a reference back to this discussion on the issue to track that we should refactor the tests in that issue as well.

d-Pixie:
Go for it! :)

PR #55, opened 2016-03-18
Implements search on repositroy base

### LINKED ISSUE #43: Search among Fortnox resources
Many Fortnox resource attributes is "searchable". It would be great to implement this functionality in our gem.

More about [Fortnox search](http://developer.fortnox.se/documentation/general/parameters/#Search).

d-Pixie:
Fixes #43


spec/fortnox/api/repositories/customer_spec.rb
```diff
@@ -4,13 +4,16 @@
 require 'fortnox/api/repositories/examples/all'
 require 'fortnox/api/repositories/examples/find'
 require 'fortnox/api/repositories/examples/save'
+require 'fortnox/api/repositories/examples/search'
 
-describe Fortnox::API::Repository::Customer do
+describe Fortnox::API::Repository::Customer, order: :defined, integration: true do
```
ehannes: Great with the `order: :defined`!

ehannes: And the `integration: true`!

spec/fortnox/api/repositories/examples/search.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      context "with one match" do
+        subject do
+          VCR.use_cassette( "#{vcr_dir}/search_by_name" ) do
+            repository.search( attribute_hash_key_name => value )
+          end
+        end
+
+        it{ is_expected.to be_instance_of( Array) }
+        it{ is_expected.to have(1).entries }
```
ehannes: Should we verify that the returned entry has the attribute that we searched for?

ehannes:
Looking great!

ehannes:
How do we add filters and searchable attributes to the repositories?

spec/fortnox/api/repositories/examples/search.rb
d-Pixie: No, that logic is not ours, so there is no point in verifying it.

d-Pixie:
Right now we allow any for both search and filter. I'll add whitelisting next.

PR #56, opened 2016-03-18
Adds filtering

### LINKED ISSUE #44: Filter Fortnox resources
Some (not all) Fortnox resources have filters. For instance, `Invoice` has filters `cancelled`, `fullypaid`, `unpaid`, `unpaidoverdue` and `unbooked`.

More about [Fortnox filters](http://developer.fortnox.se/documentation/general/parameters/#Filter).

d-Pixie:
Fixes #44 ... Sorry, too late for nice PR messages :)


d-Pixie:
Ah f**k you Code Climate :(

Any thoughts on how to fix that @ehannes ? I'm inclined to merge with it :)

ehannes:
I have a question regarding `subject` in specs. When dealing with unit testing, I have always tried to keep the `subject` to an instance of the tested unit (that is also the default value for RSpec if no `subject` is declared). If we are testing `Fortnox::API::Repository::Customer`, I try to keep the subject to an instance of that class. I see that in `spec/fortnox/api/repositories/examples/only.rb`, the `subject` is the response of a request. Whats your opinion about this? Does it make the spec more difficult to read when the subject can be different things? Does it matter at all? I really don't have a strong opinion about this, it's just how I have been working in the past :)

spec/fortnox/api/repositories/examples/only.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+      end
+
+      it{ is_expected.to be_instance_of( Array ) }
+      it{ is_expected.to have(1).entries }
+    end
+
+    context "with invalid filter" do
+      subject do
+        ->{
```
ehannes: Instead of a lambda function here, I would declare the `subject` as usual and instead send a block to `except`, like this:

```
expect{ subject }.to
```

d-Pixie: That is not possible since the `raise` I'm trying to test actually occurs when the subject is created. What you could do is stick what is in the proc into the `expect` block instead, but that kind of ruins the symmetry of the tests and makes it more difficult to read since this one will be formatted differently for the others ...

spec/fortnox/api/repositories/examples/only.rb
```diff
@@ -0,0 +1,42 @@
+
+shared_examples_for '.only' do |matching_filter, missing_filter|
+  describe '.only' do
+
+    let( :repository ){ described_class.new }
+    let( :vcr_dir ){ repository.options.json_collection_wrapper.downcase }
```
ehannes: Nice!

d-Pixie:
Re `subject` in specs it is true that the default for `subject` is an instance of the class under test, which is reasonable when you consider what subject denotes:

The subject in a test is the thing you are testing, that which you assert something of. Whatever you would out in your `expect` block is the subject of that test. In the cases where you get a default instance of your class and do something like `expect{ subject.some_method }.to` the actual subject of the test is the result of the `some_method` not the class instance.

If you are to use the one-liner syntax, `it{ is_expected.to }`, you have to do some changes to the above for it to work. Rspec also supports named subjects: `subject( :some_result ){ model.some_method }` that you can then use in your examples either in the one-liner syntax or explicitly, `expect{ some_result }.to`.

If you want to we could use the named version of `subject` whenever we are not referring to the instance of the class, but this will be most of the time anyway so the question is if it is worth it?

To me it's ok that the subject is whatever, you are declaring it locally with the tests, so it's easy enough to look up what is under test if it's not clear by the general context.

spec/fortnox/api/repositories/examples/only.rb
d-Pixie: Had to, the previous version used `subject`, but subject was declared below and they got into an infinite loop :)

d-Pixie:
Added the `while_performin` helper method.

ehannes:
I like your thoughts on `subject`. Lets go with that definition!

spec/fortnox/api/repositories/examples/only.rb
ehannes: I meant that you fetch the `vcr_dir` from `repository.options.json_collection_wrapper.downcase` :)

ehannes:
My only idea about the duplication is to extract the `VCR.use_cassette` to a method, like `vcr_cassette` which takes the block `repository.only( matching_filter )` and the cassette `"filter_hit"`. But I don't know if I like the idea myself. Maybe we do not need the duplicator engine in the specs after all... I mean, some code duplication may actually make the specs easier to read...

```
    context "with one match" do
      subject do
        VCR.use_cassette( "#{vcr_dir}/filter_hit" ) do
          repository.only( matching_filter )
        end
```

ehannes:
This PR only includes specs. I guess the code for implementing the filtering is in #55?

d-Pixie:
The implementation was actually already there :)

PR #57, opened 2016-03-18
Explicit setting of private attributes

### LINKED ISSUE #50: Add explicit writer for private attributes on model load from remote server.
# When loading resources from Fortnox, some attributes are null

I tried to fetch all `Invoice`s from Fortnox. Some of my `Invoice`s had attribute `Total` set to `100` or similar. `Total` is one of those attributes returned when fetching all `Invoice`s from Fortnox (not all are, see #25). When the gem turns the response to an entity, `invoice.total` returns `nil`.

Doing a curl request returns the correct `Total`:

```
{"@url":"https:\/\/api.fortnox.se\/3\/invoices\/5","Balance":100,"Booked":false,"Cancelled":false,"Currency":"SEK","CurrencyRate":"1","CurrencyUnit":1,"CustomerName":"Updated customer","CustomerNumber":"1","DocumentNumber":"5","DueDate":"2016-03-16","ExternalInvoiceReference1":"","ExternalInvoiceReference2":"","InvoiceDate":"2016-03-16","NoxFinans":false,"OCR":"539","WayOfDelivery":"","TermsOfPayment":"0","Project":"","Sent":false,"Total":100}
```

Even the response that the gem is getting from Fortnox (I recorded the answer via VCR in invoices/all.yml) includes the correct `Total`:

```
{"@url":"https:\/\/api.fortnox.se\/3\/invoices\/5","Balance":100,"Booked":false,"Cancelled":false,"Currency":"SEK","CurrencyRate":"1","CurrencyUnit":1,"CustomerName":"Updated customer","CustomerNumber":"1","DocumentNumber":"5","DueDate":"2016-03-16","ExternalInvoiceReference1":"","ExternalInvoiceReference2":"","InvoiceDate":"2016-03-16","NoxFinans":false,"OCR":"539","WayOfDelivery":"","TermsOfPayment":"0","Project":"","Sent":false,"Total":100}
```

The same goes for `balance`, `booked`, `cancelled` and maybe other attributes as well. The attributes that seems to be working are `customer_number`, `customer_name`, `due_date` and `currency`, `currency_rate`, `currency_unit`, `document_number`, `ocr`.

d-Pixie:
Adds explicit writers for private attributes when the model is created from a repository.

Fixes #50


d-Pixie:
Ok @ehannes this is what I know: The deep root of the error is inside Virtus cohersion when it tries to set a new cohersion for a specific attributes class. The actual problem, I'm almost certain, is that the shared example that is focused does something weird that reuses state in a way that shouldn't be possible.

The focused example runs two tests, the order of them is random and the first on always succeeds and the second always fails. Leading me to think it's state related :)

If you comment out the `initialize_private_attributes` call in the model base constructor the error disappears, why I have no clue. The logic in that method shouldn't affect anything like this, but apparently it does :/

It's a mess and I just feel that if this was an actual bug in our code we should hit it more than once in ~700 tests ... But maybe not. Have a look and see what you think :)

Time to sleep, I have to get up early tomorrow on top of everything :zzz:

ehannes:
Is this the same error you get?

```
  1) Fortnox::API::Repository::Invoice.save old (update existing) include correct :comments value
     Failure/Error: super hash
     
     RuntimeError:
       can't modify frozen Hash
```

ehannes:
Do you know why the same test succeeds in `fortnox/api/repositories/customer_spec`?

ehannes:
Why should the spec succeed if we commented out `initialize_private_attributes`? That seemed weird since it fails on `super hash`, which is called before the previously mentioned method. Thats why I though: maybe the `initialize` method is run more than once. In fact, when I run the focused test, the `initialize` method in `Model::Base` is run 4 times. Shouldn't it be two times? One for the find request and one for the save...

d-Pixie:
Yes, it's the second run of the main initialize method that is the problem I think. Have no idea why it's happening in the first place. If it's a circular dep. it should be an infinite loop, not just twice :/

But, as we discussed on Slack, we'll probably get rid of this when migrating away from Virtus towards dry-validation.

ehannes:
Lets hope so!

ehannes:
@d-Pixie we still need to fix readonly-attributes, but I guess we should do that in a new branch now that we have implemented dry-types? Or are there parts here that you want to reuse?

d-Pixie:
With the changes done in #73 this situation has changed and I'll open a new PR for an implementation of this with the new code.

PR #58, opened 2016-04-20
Order

ehannes:
New PR to add `Order` to the gem. (Old PR #13 was opened to `master` and not `development`.)

`Order` and `Invoice` has a lot in common. The shared parts are placed in a module called `OrderBase` (a better module name would be great :)).


ehannes:
The current solution for adding attributes in `OrderBase` is probably bad. Code Climate does not like it at all (_Method has too many lines. [56/30]_). Do you have any thought on this @d-Pixie?

d-Pixie:
A better module name is OOIBase (order, offer, invoice) I think, or to copy Fortnox internally; DocumentBase.

d-Pixie:
Regarding the CC complaint I think we should leave it. We had the same problem before and while we could refactor this into several smaller methods it doesn't really make sense here. It's one context and we are writing framework code, so we are more likely to get into edge cases. I also think that moving it into smaller methods would reduce the readability, not increase it, so no, just leave it.

Possibly adding an exception for the file in the CC config (there is a link on CC above the error that shows you how to do it).

ehannes:
`OOIBase` is not bad if you know what those letters stands for. If you don't, you have no idea of what that module does. `DocumentBase` may not be the best name either, but at least it gives you a clue in the right direction.

d-Pixie:
We could always adda docblock to the file and explain it :)

ehannes:
We could add a docblock, but I still think `DocumentBase` is a clearer. The module includes a `document_number` attribute, suggesting that it actually is a document. And I have replaced all references to `Invoice` or `Order` in the attribute comments to `Document` which I think make sense. For instance `# Freight Freight cost of the invoice` is now `# Freight Freight cost of the document`.

d-Pixie:
Yeah, and generally the end user of the library will not be concerned with this detail anyway so ...

lib/fortnox/api/models/invoice.rb
```diff
@@ -1,142 +1,43 @@
-require "fortnox/api/models/base"
+require "fortnox/api/models/document_base"
 require "fortnox/api/models/edi_information"
-require "fortnox/api/models/email_information"
-require "fortnox/api/models/row"
+require "fortnox/api/models/invoice_row"
```
d-Pixie: Question on this: is the invoice row and the other document rows different? If they are the same it should probably have some more generic name ...

lib/fortnox/api/models/invoice_row.rb
```diff
@@ -0,0 +1,15 @@
+require "fortnox/api/models/row"
+
+module Fortnox
+  module API
+    module Model
+      class InvoiceRow < Row
```
d-Pixie: Ah, ok, it is a specialisation :)

d-Pixie:
Looks good to me. Go ahead and merge it whenever you are ready.

lib/fortnox/api/models/invoice.rb
ehannes: We have `Fortnox::API::Model::Row` as a superclass for `Rows`. Maybe we should rename it to `DocumentRow` now? Then we have both `InvoiceRow` and `OrderRow` that inherits from `Row`. Maybe we should get rid of this inheritance also and use a module again?

d-Pixie: Well, if we are not going to have any other row models we can leave it as just `Row`. About the module argument it depends. If this is specialisations it's fine to use inheritance, but if it's just that they share some properties (which I think is the case here) we should change it into a mixin instead.

ehannes:
I'll need to implement the `Fortnox::API::Repositoriy::Order` and `Fortnox::API::Validator::Order` first.

lib/fortnox/api/models/invoice.rb
ehannes: They only share attributes. I'll change it to a mixin instead :) And about the name: I do not think there will be any other `Row`s, but if I use mixin, I think `DocumentRow` will be a lot clearer than `Row`, don't you think?

d-Pixie: Either is fine, if you like DocumentRow better it is fine by me :)

ehannes:
@d-Pixie: I think this PR is ready for some code review now :)

ehannes:
About CC issue: I vote for an exception for that "Similar code found in 1 other location"-issue.

lib/fortnox/api/models/document_base.rb
```diff
@@ -0,0 +1,183 @@
+require "fortnox/api/models/base"
+require "fortnox/api/models/email_information"
+
+module Fortnox
+  module API
+    module Model
+      module DocumentBase
+        #rubocop:disable Metrics/MethodLength
+        def self.included(base)
+          # Url Direct url to the record.
```
d-Pixie: I would add a comment here to document why we decided to do it this way instead of breaking it into several smaller methods. The readability argument in other words.

lib/fortnox/api/models/document_row.rb
```diff
@@ -0,0 +1,79 @@
...
+      module DocumentRow
+        DISCOUNT_TYPES = %w(AMOUNT PERCENT)
+        HOUSE_WORK_TYPES = %w(CONSTRUCTION ELECTRICITY GLASSMETALWORK GROUNDDRAINAGEWORK MASONRY PAINTINGWALLPAPERING HVAC CLEANING TEXTILECLOTHING COOKING SNOWPLOWING GARDENING BABYSITTING OTHERCARE TUTORING OTHERCOSTS)
+
+        def discount_type=( raw_discount_type )
+          discount_type = raw_discount_type.upcase
+          super discount_type if DISCOUNT_TYPES.include? discount_type
+        end
+
+        def house_work_type=( raw_house_work_type )
```
d-Pixie: Look at the attributes for these. If you need to do data mangling it's probably worth making it into an attribute. But skip it for now, the attributes implementation will be a bit different using dry-types...

lib/fortnox/api/repositories/base/options.rb
```diff
@@ -8,6 +8,7 @@ class Options
                         :unique_id, :attr_to_json_map, :json_to_attr_map,
                         :keys_filtered_on_save
 
+          #rubocop:disable Metrics/ParameterLists
```
d-Pixie: Yeah ... That one is a bit of a problem. I guess I could turn it into a DSL instead of a massive constructor. But a lot of this is required anyway, or will be defined for each modell at least. I'll think about it.

lib/fortnox/api/repositories/order.rb
```diff
@@ -0,0 +1,33 @@
+require "fortnox/api/repositories/base"
+require "fortnox/api/models/order"
+
+module Fortnox
+  module API
+    module Repository
+      class Order < Fortnox::API::Repository::Base
+
+        CONFIGURATION = Fortnox::API::Repository::Base::Options.new(
```
d-Pixie: So this could be something like:

    Fortnox::API::Repository::Base::Options.new do
      uri '/orders/',
      json_collection_wrapper 'Orders',
      json_entity_wrapper 'Order',
      unique_id 'DocumentNumber',
      attribute_name_to_json_key_map { administration_fee_vat: 'AdministrationFeeVAT', edi_information: 'EDIInformation', freight_vat: 'FreightVAT', total_vat: 'TotalVAT', vat_included: 'VATIncluded' }
      keys_filtered_on_save [ :url ]
    end

Which would silence the warning about too many arguments. But would it be better or as good?

d-Pixie:
Nothing major, more thoughts than anything to fix. Looks good to me.

lib/fortnox/api/models/document_row.rb
ehannes: Added a low priority issue for that: #61

lib/fortnox/api/repositories/base/options.rb
ehannes: Yep. If we decide to do something about it, that would be suitable for a separate PR.

lib/fortnox/api/repositories/order.rb
ehannes: As I wrote above, maybe something for a separate PR. This concerns all `Repositories`, not just `Order`.

PR #63, opened 2016-04-27
Fix 62

### LINKED ISSUE #62: Creating model with nested models including nil values
When creating a model, attributes not provided will have nil as value. When corresponding repository tries to save the model, it should remove nil values. `Repository::JSONConversion.remove_nil_values( hash )` manages this, but it does not handle nested hashes yet.

It is defined as
```
def remove_nil_values( hash )
  hash.delete_if{ |_, value| value.nil? }
end
```

Ex:
```
irb(main):054:0> order = {:order=>{:order_rows=>{"1"=>{"price"=>0, "ordered_quantity"=>0, "article_numer"=>"0000", "bar"=>nil}}}, :foo=>nil}
=> {:order=>{:order_rows=>{"1"=>{"price"=>0, "ordered_quantity"=>0, "article_numer"=>"0000", "bar"=>nil}}}, :foo=>nil}
irb(main):055:0> order.delete_if{ |_, value| value.nil? }
=> {:order=>{:order_rows=>{"1"=>{"price"=>0, "ordered_quantity"=>0, "article_numer"=>"0000", "bar"=>nil}}}}
```
`:foo` is correctly removed, but `:bar` is not.

ehannes:
Fixes #62. It introduces nested translation hashes for nested attributes. It only includes an implementation for `Order` and its nested `OrderRows`.

The solution is not perfect. Right now, the nested model is defined as a nested hash in the repository, for instance:

```
 attribute_name_to_json_key_map: {
  administration_fee_vat: 'AdministrationFeeVAT',
  edi_information: 'EDIInformation',
  freight_vat: 'FreightVAT',
  total_vat: 'TotalVAT',
  vat_included: 'VATIncluded',
  order_rows: {
    vat: 'VAT'
  }
},
```

Here, `order_rows` is a nested hash. But what if the attribute `order_rows` itself needs a translation? Additionally, what if the nested model's translation does not need any specific JSON translations? Would you then implement it as just an empty hash? Somehow `JSONConvertion` must know that an attribute is a nested model. Otherwise, it will not recursively translate that nested model's attributes.

A better solution may be to add a new hash to the Repository named `nested_models` where you declare what nested models there are. Then you only add them to `attribute_name_to_json_key_map` if they have some attribute that needs a specific translation.

### DIFF
--- a/lib/fortnox/api/models/order.rb
+++ b/lib/fortnox/api/models/order.rb
@@ -19,6 +19,13 @@ module Fortnox
 
         # OrderRows Separate object
         attribute :order_rows, Array[OrderRow]
+
+        # Override to_hash to convert nested OrderRows to hash correctly.
+        def to_hash *args
+          main_hash = super
+          main_hash[:order_rows] = self.order_rows.map(&:to_hash)
+          main_hash
+        end
       end
     end
   end
--- a/lib/fortnox/api/models/order.rb
+++ b/lib/fortnox/api/models/order.rb
@@ -19,13 +19,6 @@ module Fortnox
 
         # OrderRows Separate object
         attribute :order_rows, Array[OrderRow]
-
-        # Override to_hash to convert nested OrderRows to hash correctly.
-        def to_hash *args
-          main_hash = super
-          main_hash[:order_rows] = self.order_rows.map(&:to_hash)
-          main_hash
-        end
       end
     end
   end
--- a/lib/fortnox/api/repositories/base/json_convertion.rb
+++ b/lib/fortnox/api/repositories/base/json_convertion.rb
@@ -6,14 +6,14 @@ module Fortnox
       module_function
 
         def hash_to_entity( entity_json_hash, key_map )
-          remove_nil_values(entity_json_hash)
+          entity_json_hash = clean_hash(entity_json_hash)
           entity_hash = convert_hash_keys_from_json_format( entity_json_hash, key_map )
           instansiate( entity_hash )
         end
 
         def entity_to_hash( entity, key_map, json_entity_wrapper, keys_to_filter )
           entity_hash = entity.to_hash
-          clean_entity_hash = remove_nil_values( sanitise( entity_hash, keys_to_filter ) )
+          clean_entity_hash = clean_hash( sanitise( entity_hash, keys_to_filter ) )
           entity_json_hash = convert_hash_keys_to_json_format( clean_entity_hash, key_map )
           { json_entity_wrapper => entity_json_hash }
         end
@@ -58,20 +58,15 @@ module Fortnox
           end
         end
 
-        # Removes nil values recursively from Hash
-        def remove_nil_values( hash )
-          hash.each do |key, value|
-            if value.nil?
-              hash.delete( key )
-            else
-              clean_collection( value )
-            end
+        # Removes nil values and empty collections recursively from Hash
+        def clean_hash( hash )
+          cleaned_hash = hash.select{ |k, v| v != nil }
+          cleaned_hash.each do |k, v|
+            cleaned_hash[k] = clean_collection( v ) if v.respond_to?( :each )
           end
-
-          hash.delete_if{ |_, v| v.empty? if v.respond_to?( :empty? ) }
         end
 
-        def clean_collection ( collection )
+        def clean_collection( collection )
           case collection
           when Array
             collection = clean_array( collection )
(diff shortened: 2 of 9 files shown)


ehannes:
@d-Pixie Any thought on the code duplication that CC found?

ehannes:
Travis reports that Ruby 2.1.8 does not like our gem any more. Why does it try to use `listen 3.1.1` now? This PR does not add or change any dependencies. Maybe some of the gems we are using has updated their dependencies?
```
Installing listen 3.1.1
Gem::InstallError: listen requires Ruby version ~> 2.2.
Installing vanguard 0.0.5
An error occurred while installing listen (3.1.1), and Bundler cannot continue.
Make sure that `gem install listen -v '3.1.1'` succeeds before bundling.
```

ehannes:
From [listen's GitHub page](https://github.com/guard/listen#important-ruby-21-is-officially-outdated-and-unsupported-please-upgrade-to-ruby-22-before-installing-listen): _IMPORTANT: Ruby 2.1 is officially outdated and unsupported! Please upgrade to Ruby 2.2 before installing Listen!_

So let's simply drop support for Ruby 2.1.

ehannes:
@d-Pixie When you have time, take a look at this. (No hurry).

d-Pixie:
"What if the attribute itself needs a translation?" - Do we have this case for any models? If not why solve it? :)

"What if the nested model's translation does not need any specific JSON translations? Would you then implement it as just an empty hash? Somehow JSONConvertion must know that an attribute is a nested model. Otherwise, it will not recursively translate that nested model's attributes." - This is self contradictory. Either you need attributes in the hash, since the attributes needs translating, or not. If you don't need them it's easy to translate the parent key, it will be like any other key.

The problem only occurs if you need to translate both the parent key AND attributes. We could add a special key for this, `_self_` or `_parent_` or something. Or we could make the value an array `[ 'parent_key_translation', { child: 'key' }]`. Or we could come up with something different completely.

lib/fortnox/api/models/order.rb
```diff
@@ -19,6 +19,13 @@ class Order < Fortnox::API::Model::Base
 
         # OrderRows Separate object
         attribute :order_rows, Array[OrderRow]
+
+        # Override to_hash to convert nested OrderRows to hash correctly.
+        def to_hash *args
```
d-Pixie: Did you try without this? In my would to_hash should be recursive by default, but I might well be wrong...

ehannes:
We have the following case:
An `Invoice` has one `EDIInformation` which has nested attributes like  `EDIGlobalLocationNumber`.

lib/fortnox/api/repositories/base/json_convertion.rb
```diff
@@ -39,7 +39,16 @@ def default_key_from_json_transform( key )
 
         def convert_hash_keys_to_json_format( hash, key_map )
           hash.each_with_object( {} ) do |(key, value), json_hash|
-            json_hash[ convert_key_to_json( key, key_map ) ] = value
+            if value.is_a?(Array) # Is attribute a nested model?
+              nested_models = []
+              value.each do |nested_model|
+                nested_key_map = key_map.fetch( key, {} )
```
d-Pixie: What if you get a string back here, as in the parent key needs translating but not the nested model attributes?

Wouldn't it be better to base the recursion on how the  translations map looks? If it has a hash as the value for this key you know you need to recurse.

Right now you will error out if: 1) The parent key needs a translation but not the nested model. 2) The array is of atomic values and not objects, such as several ENUM values etc. 3) The nested model is singular, ie. not an array.

lib/fortnox/api/models/order.rb
ehannes: I did. But feel free to give it another try :) Just remove that method and see if the tests passes.

lib/fortnox/api/repositories/base/json_convertion.rb
```diff
@@ -58,10 +67,32 @@ def sanitise( hash, keys_to_filter )
...
+            collection = clean_hash( collection )
+          end
+
+          collection
+        end
+
+        # Removes nil values and empty collections recursively from Array
+        def clean_array( array )
+          cleaned_array = array.select{ |e| e != nil }
+          cleaned_array.each_with_index do |e, i|
```
d-Pixie:
```
cleaned_array.map{|e| e.respond_to?( :each ) ? clean_collection( e ) : e }`
```
Or even better, make `clean_collection` more general to `clean_value` and just add a check like this to the top of it:
```
return value unless value.respond_to?( :each )
```
That way you get this here instead:
```
cleaned_array.map{|e| clean_value( e )}`
```

d-Pixie:
I would suggest that you implement the `to_hash` method on `Model::Base` instead. That is inherited by everyone and you can then walk arrays and check if any of the values respond to `.to_hash` in your implementation instead of having to reimplement `to_hash` in every model that needs it.

lib/fortnox/api/models/order.rb
ehannes: I just gave it another try and we do need the `to_hash` method :( Without that you get something like this:

```
Fortnox::API::Model::Order.new(order_rows: Fortnox::API::Model::OrderRow.new).to_hash
# => {..., :order_rows=>[#<Fortnox::API::Model::OrderRow:0x007f0dc3fc77e0 @saved=false, @new=true, @account_number=nil, @article_number=nil, @contribution_percent=nil, @contribution_value=nil, @cost_center=nil, @delivered_quantity=nil, @description=nil, @discount=nil, @discount_type=nil, @house_work=nil, @house_work_hours_to_report=nil, @house_work_type=nil, @price=nil, @project=nil, @total=nil, @unit=nil, @vat=nil, @ordered_quantity=nil>]}
```

ehannes:
As you suggested, a99817a  moves the implementation of `to_hash` to`Model::Base`.

ehannes:
This PR will be replaced with a new one instead... It's to old and we will use a different approach now that `Mapper`s is implemented.

PR #65, opened 2016-04-28
Tests for Repository::Order added

ehannes:
Spec for Repository::Order now includes both shared examples groups
'.search' and '.only'. These groups had to be refactored to
make it work. For instance, there is no filter for Order that
returns no hit, so that test was omitted for Order.


ehannes:
@d-Pixie I'll merge this directly. It only concerns specs. If you have any objections, I'll fix it afterwards :)

PR #73, opened 2016-08-29
Introduce dry types

### LINKED ISSUE #35: Coverage validator
We need to add a validator that uses `cover?`instead of `include?` since we have some really large numbers that we test against ...

### LINKED ISSUE #36: Named imported attributes
When we use attribute composition in models today we declare the name of the attribute in the attribute declaration itself, not when using the attribute in a model. For things like account this won't work since accounts exist in many different models under different names.

We need to find a nice way of declaring the name of the included attribute at include time instead of at declaration time.

### LINKED ISSUE #50: Add explicit writer for private attributes on model load from remote server.

### LINKED ISSUE #59: Nested validators
Some models have nested models. For instance `Fortnox::API::Model::Invoice` has `Fortnox::API::Model::InvoiceRow`, `Fortnox::API::Model::EmailInformation` and `Fortnox::API::Model::EDIInformation`. That means `Fortnox::API::Validator::Invoice` should also validate nested models.

### LINKED ISSUE #61: DiscountType and HouseWorkType attributes
`Fortnox::API::Model::DocumentRow` has some logic for `discount_type` and `house_work_type`. It would be better to make separate attribute classes instead.

### LINKED ISSUE #66: Limits for DocumentBase validator
I checked with Fortnox and apparently their documentation for is incorrect. These are the limits that Fortnox have given me:

### Order and Invoice
administration_fee:	_None_
currency_rate: _None_
currency_unit: _None_
freight: 9999999999.00
address1: 1024
address2: 1024
customer_name: 1024
external_invoice_reference1: 80
external_invoice_reference2: 80
our_reference: 50
phone1: 1024
phone2:	1024
remarks: 1024
your_order_number: 30
your_reference:	50
zip_code: 1024

#### Note about None
According to Fortnox support:
> eftersom det är ett numeriskt värde så borde det inte vara någon begränsning

### Customer
DeliveryAddress1: 1024
DeliveryAddress2: 1024
DeliveryCity: 1024
Comment: 1024
DeliveryZipCode: 1024
City: 1024

### LINKED ISSUE #72: dry types
Since we have found some cases where Virtus doesn't quite cover us and dry types is more explicitly what we need we are switching to dry types for our domain model implementation. This will also fix #62 and unblock #64...

### LINKED ISSUE #74: Support for multiple access tokens
Fortnox has a [request limits on their API](http://developer.fortnox.se/blog/important-implementation-of-rate-limits/). This limit (16/second, 1000/minute) could easily be reached if you do bulk edit or many users. Fortnox suggests using several Access Tokens to solve this. Would it be possible for this gem to handle such a feature?

Fortnox returns status `503` when this limit is reached. Could the repository classes in this gem look for a `503` status and switch Access Token automatically?

ehannes:
Fixes #35
Fixes #36 
Fixes #50 
Fixes #59 
Fixes #61
Fixes #66 
Fixes #72 
Fixes #74


ehannes:
@d-Pixie Should `test.rb` be removed before merge?

ehannes:
If I use this gem in an application and try to create a model from user input which gives a `ConstraintError` like the one below, how can I tell my users what went wrong?
```
Dry::Types::ConstraintError:
       "PRIVATEMORE STRING" violates constraints (included_in?(["PRIVATE", "COMPANY"]) failed)
```

ehannes:
What does `.with( read_only: true )` do with our models? I think the wanted behavior in our case is to not be able to set that attribute when initializing the model. I was able to initialize an `Order` with `total_vat`, without any `Dry::Types::ConstraintError`. When I tried to save the order, I got the following error from Fortnox:

```
     Fortnox::API::RemoteServerError:
       Fältet TotalVAT är endast läsbart.
```

ehannes:
About `.with( read_only: true )`:
Read only attributes is not working in current implementation of the gem (see #50). This PR introduces dry-types, and meta information about which attributes should be read only is included, but it does not include the logic to actually make them read only. Let's leave that to a separate PR.

ehannes:
## A notation about enums

`CountryCode` and `Currency` automatically truncate given value to 2 and 3 characters. For instance, `Currency` accepts `SEKMORE STRING` (will be truncated to `SEK`). The other enums does not behave like this. For instance,  `VATTypes` raises `ConstraintErrir` for `EUVATMORE STRING`

I guess this is because of `EnumConstructors.sized(2)` and `EnumConstructors.sized(3)` in `lib/fortnox/api/types.rb`.

ehannes:
I tried to inspect a `Dry::Types::ConstraintError`, and this is what i got:
```
#<Dry::Types::ConstraintError: "PRIVATESTRING" violates constraints (included_in?(["SEVAT", "SEREVERSEDVAT", "EUREVERSEDVAT", "EUVAT", "EXPORT"]) failed)>
```
So there is no information about what attribute made the violation. [Link to source code](https://github.com/dry-rb/dry-types/blob/adcf3587a52c6b694744b09e67d0cf997acd2574/lib/dry/types/errors.rb). So we should probably see these `ConstraintError`s as developer errors, and not use them for feedback to the end user. The question is then how the gem should be used: What about if an end user gives the gem invalid data?

ehannes:
Before merging this PR, we should probably remove all VCR cassettes and record them again, just to make sure all requests are recorded correctly. VCR sometimes raises an error when you add data to your request, but I don't know how sensitive it is. Recently, I have changed some requests recorded by VCR and no errors were raised. I was forced to remove the cassette manually.

ehannes:
Shouldn't this code raise a `Dry::Types::StructError` or similar, or am I missing something here?
```
irb(main):013:0> Fortnox::API::Model::Customer.new(name: 'Kurt Nilsson',
                                                   type: 'INVALID CUSTOMER TYPE')                                                                                                       
=> #<Fortnox::API::Model::Customer url=nil ... type="INVALID CUSTOMER TYPE" ...>
```



`:type` is defined as `attribute :type, Fortnox::API::Types::CustomerType`.

d-Pixie:
About the country code and currency enums subtypes. These are special since they are always 2 and 3 characters long respectively. The other enums does not conform in that way and so a similar aid is not possible for them.

If you feel that this is inconsequent we can remove the behaviour. But they are not strictly enum types, they are just a custom type that happen to be based on enum. But that is an implementation detail and to me it made sense that they would try to help the user as much as possible...

ehannes:
Then it works as expected. The tests are adapted to this behavior (with `auto_crop:true`), so I think we can leave it as so.

ehannes:
I realize now that test coverage for models is really low (see examples with `Customer` above for instance). We moved logic from our validators into our models and instead of moving the tests along, we simply deleted them. I'll reintroduce them (shouldn't take that long).

ehannes:
Wow, nice @d-Pixie. Is that a new feature, to "approve changes"?

d-Pixie:
Yeah, I think it came in together with the projects feature. Something to improve the PR process :)

Maybe you should add simplecov as well to guard the coverage in the future?

I don't see any reason not to merge this when you are happy with the test coverage. We should probably refactor some more down the line but some actual use will probably help in guiding those refactorings so let us leave it as is for now.

ehannes:
That's great news!
I'm working on model tests right now. I'll try to include simplecov.

ehannes:
Unfortunately, this PR will mess up the code a bit. It has grown to be a huge refactoring and it will introduce some new problems (for instance issue #75) to fix before we can release this code. Also, Code Climate refuses to look at this PR any more since it compares this PR against a commit on `development`  that I apparently force pushed (something  CC does not like). But anyway, we really need to merge this now, so I'll do it anyway and we'll fix the problems afterwards.

PR #76, opened 2016-10-05
Introduce rubocop rspec

ehannes:
Introduces rubocop-rspec, a Rubocop engine for RSpec.

We recently merged a branch that had `focus: true` set on one test, which meant only that test in the whole test suite was run... This sadly lead to a merge with failing code to `development` branch. rubocop-rspec will check things like this for us :) Unfortunately, this means that tests will not pass in this branch, since `development` runs failing code. In my opinion, the failing tests should not be fixed in this branch. This should only include rubocop-rspec to our project. I'll vote for dealing with the failing specs in another branch.

This commit includes a lot of fixes for Rubocop violations among our specs.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -988,6 +988,11 @@ Style/SpaceInsideRangeLiteral:
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-space-inside-range-literals'
   Enabled: false
 
+Style/SpaceInsideStringInterpolation:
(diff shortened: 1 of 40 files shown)


ehannes:
Please merge #80 before merging this. #80 fixes spec failures that this branch inherited from `development` branch.

ehannes:
@d-Pixie: Could you have a look at the CC issues? I've solved all except the two ones that are left.

PR #80, opened 2016-10-05
Fixes bug in Fortnox::API::Base

ehannes:
Fixes bug in Fortnox::API::Base

Focused test in Fortnox::API::Base spec is removed.
When the focus was removed, three tests in Fortnox::API::Base spec was failing. I traced the bug to line 40 in Fortnox::API::Base. For some reason, response should not be returned here, or else no Exceptions are raised on remote server errors (500 response from Fortnox). I'll let @d-pixie have a look at this before merge...

### DIFF
--- a/lib/fortnox/api/base.rb
+++ b/lib/fortnox/api/base.rb
@@ -37,7 +37,7 @@ module Fortnox
           execute do |remote|
             response = remote.send( method, *args )
             self.headers['Access-Token'] = get_access_token
-            return response
+            response
           end
         end
       end
--- a/spec/fortnox/api/base_spec.rb
+++ b/spec/fortnox/api/base_spec.rb
@@ -96,15 +96,15 @@ describe Fortnox::API::Base do
       )
     end
 
-    it 'works', focus: true do
+    it 'works' do
       api = Fortnox::API::Base.new
       response1 = api.get( '/test', { body: '' })
       response2 = api.get( '/test', { body: '' })
       response3 = api.get( '/test', { body: '' })
 
-      expect( response1.parsed_response ).to_not eq( response2.parsed_response )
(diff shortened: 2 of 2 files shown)


lib/fortnox/api/base.rb
```diff
@@ -37,7 +37,7 @@ def initialize
           execute do |remote|
             response = remote.send( method, *args )
             self.headers['Access-Token'] = get_access_token
-            return response
```
d-Pixie: Ah yes ... Silly mistake, since the return is in a block it will return not only the block but the entire function the block is in :)

spec/fortnox/api/base_spec.rb
```diff
@@ -96,15 +96,15 @@
...
       api = Fortnox::API::Base.new
       response1 = api.get( '/test', { body: '' })
       response2 = api.get( '/test', { body: '' })
       response3 = api.get( '/test', { body: '' })
 
-      expect( response1.parsed_response ).to_not eq( response2.parsed_response )
-      expect( response3.parsed_response ).to_not eq( response2.parsed_response )
-      expect( response1.parsed_response ).to eq( response3.parsed_response )
+      expect( response1 ).to_not eq( response2 )
+      expect( response3 ).to_not eq( response2 )
```
d-Pixie: Hmm, are you sure this works? I had it like that first but the test errored out even when the printed objects in the fail message were identical...

But if it works it's better this way.

ehannes: As you wrote above:
> since the return is in a block it will return not only the block but the entire function the block is in...
The entire function was returned before. That's why you had to run `parsed_response`. Now, when the entire function is no longer returned but only the actual response this is the correct implementation.

PR #81, opened 2016-10-05
Introduces mappers

ehannes:
A simple implementation of mappers.

Mappers extract logic from Repository::Base::Options that concerns mapping between Fortnox JSON-data and our data models. This branch does not include any new features. When this PR is merged, we can move on and solve our mapping-related problems.

Please merge #80 first.


ehannes:
We have some code duplication in our repository classes. Is it safe to move from:
```
module Fortnox
  module API
    module Repository
      class Order < Fortnox::API::Repository::Base
      # ...
      MODEL = Fortnox::API::Model::Order
      MAPPER = Fortnox::API::Mapper::Order
      # ...
    end
  end
end
```
to:
```
module Fortnox
  module API
    module Repository
      class Order < Base
      # ...
      MODEL = Model::Order # or should it be ::Model::Order?
      MAPPER = Mapper::Order # or should it be ::Mapper::Order?
      # ...
    end
  end
end
```

Do you have any other suggestions on how to reduce the duplication? Maybe the constant could be replaced by class variables in the base class (although they are constants... hmm...).

d-Pixie:
Just leave that duplication. Working around it is just to satisfy the linter. In reality that code is factory code that will go away when we move to using dry container...

PR #82, opened 2016-10-10
Nested mappers

### LINKED ISSUE #15: Fix mapping from JSON to Ruby objects for Attributes
Some model attributes needs special care when converting to/from JSON. This is currently handled in the model's repository. This functionality is needed for our `Valid Objects`/`Attributes` (placed in `lib/fortnox/api/models/attributes`) too.

### LINKED ISSUE #62: Creating model with nested models including nil values

### LINKED ISSUE #64: JSONConvertion.save should return model instance, not JSON
When saving a new instance of a Fortnox resource, the repository currently returns JSON instead of the saved model.

```
puts API::Fortnox::Repository::Order.new.save(model)
# => {"Order"=>{"@url"=>"https://api.fortnox.se/3/orders/18", "@urlTaxReductionList"=>"https://api.fortnox.se/3/taxreductions?filter=orders&referencenumber=18", "AdministrationFee"=>0, "AdministrationFeeVAT"=>0, "Address1"=>"", "Address2"=>"", "BasisTaxReduction"=>0, "Cancelled"=>false, "City"=>"", "Comments"=>"A great comment about something", ...}}
```

Maybe it should return the saved instance instead?

### LINKED ISSUE #75: Record new VCR-cassettes
We need to refresh our VCR-cassettes after the implementation of `dry-types` in our gem. I think there will be a few bugs to correct...

ehannes:
Implements nested mappers. Before, we had mappers for our models, for instance `Order`. But we did not handle nested models like an `OrderRow`. This PR includes logic to handle such nested models and maps them correctly to and from JSON.

Fixes #15
Fixes #62
Fixes #64
Fixes #75 

Opens #83


ehannes:
Right now, this PR includes a lot of messy code. But if you could look at the `JSONConversion` and `spec/fortnox/api/mappers/base/json_convertion.rb` spec, it would help a lot I think.

lib/fortnox/api/types/model.rb
```diff
@@ -6,7 +6,7 @@ class Model < Dry::Struct
 
         def initialize( input_attributes )
           if (missing_key = first_missing_required_key( input_attributes ))
-            raise KeyError.new( missing_key )
+            raise KeyError.new( "#{missing_key} in #{input_attributes}" )
```
ehannes: :+1:

ehannes:
I think the only thing left now is to get the specs working again...

ehannes:
Currently, two Order specs are failing (`Internal Error` when trying to save `Order`s). I have no clue why, but we really need them to pass before merging. I've asked Fortnox about this and their answer on Friday 16:43 was
> Jag har skickat detta till våra utvecklare som ska se över varför felmeddelandet visas.
> Vi återkommer när vi har mer information.

PR #89, opened 2016-10-25
Fix #88

### LINKED ISSUE #88: Format for AccessToken in .env file not working
To use token rotation, the gem assumes that they are separated with `\n`. The problem is that when the value is read from the `.env`-file, it is escaped like this: `xxxxxxxxx-yyyyy-zzzz-aaaaa-bbbbbbbbbbbbb\\nxxxxxxxxx-yyyyy-zzzzz-aaaa-bbbbbbbbbbbb` Notice the `\\n` separating the two tokens.

We did not get this problem when running our tests because the values from our `.env`-file are never used in our tests. Instead, we have a shared context for repositories that sets the environment variables and our `spec_helper` resets these values after each test run. We should remove this and use our `.env`-file instead since current approach is error prone.

ehannes:
Changes the delimiter used to declare multiple `AccessToken`s in environment variables (newline is changed to comma).

Tests are now using the `.env.test`-file.


spec/fortnox/api/repositories/customer_spec.rb
```diff
@@ -22,10 +19,11 @@
                    'test@example.com'
 
   # It is not yet possible to delete Customers. Therefore, expected nr of
-  # Customers when running .all will continue to increase.
+  # Customers when running .all will continue to increase
+  # (until 100, which is max by default).
   include_examples '.all', 100
```
d-Pixie: I had a thought about this before, unrelated to this PR but I'm telling you now since I remember :) :

I know you can make some tests in rspec sequential. We should change this to be `.all`, save the number of customers, `.new`, `.all` again and compare to the saved number. That way we get dynamic tests without hardcoded values that we have to update.

ehannes: Good point!

Actually, our repository specs are run sequentially with the `order: :defined` in the spec declaration.

The tricky part is that there is a max limit of how many records Fortnox will return. Default is 100, which is what we have now. If we run `.all`, we will get 100 records back. Then we create a new one and run `.all` again and we will still get 100 records back.

spec/fortnox/api_spec.rb
```diff
@@ -5,10 +5,14 @@
 
   context 'get access token' do
     before do
-      ENV['FORTNOX_API_BASE_URL'] = 'http://api.fortnox.se/3/'
-      ENV['FORTNOX_API_CLIENT_SECRET'] = 'P5K5vE3Kun'
-      ENV['FORTNOX_API_ACCESS_TOKEN'] = '3f08d038-f380-4893-94a0-a08f6e60e67a'
-      ENV['FORTNOX_API_AUTHORIZATION_CODE'] = 'ea3862b0-189c-464b-8e23-1b9702365ea1'
+      stub_const(
```
d-Pixie: A lot of these now ... How about declaring a helper that does the around stuff:

```
def stub_environment( hash )
  stub_const( 'ENV', ENV.to_hash.merge( hash ))
end
```
and just call it like this:
```
stub_environment(
  'FORTNOX_API_BASE_URL' => 'http://api.fortnox.se/3/',
  'FORTNOX_API_CLIENT_SECRET' => 'P5K5vE3Kun',
  'FORTNOX_API_ACCESS_TOKEN' => '3f08d038-f380-4893-94a0-a08f6e60e67a',
  'FORTNOX_API_AUTHORIZATION_CODE' => 'ea3862b0-189c-464b-8e23-1b9702365ea1'
)
```

spec/fortnox/api/repositories/customer_spec.rb
d-Pixie: Yes, I don't think that fix belongs here. Go for noting it for now.

PR #91, opened 2016-10-26
Fixes #87 Token rotation

### LINKED ISSUE #87: Improve token rotation
- [ ] Each time we initialize a `Repository`, token rotation is reset. Normally, a web application will initialize a new `Repository` for each application request. Rotation is using currently instance variables, so if you only do one Fortnox request per web app request, you get no rotation at all.

- [ ] Rotation starts on index 1 rather than index 0 which would be the logic starting point (if we don't implement random start index).

ehannes:
Token rotation should be done on class level, not instance level.
See #88 for more details.

Instead of setting Access-Token both when initialized and after each
request, it is not only set BEFORE each request. This is to handle both
token rotation for subsequent requests with the same repository object
as well as requests with several repository objects.

Dummy VCR cassettes added.


ehannes:
Now token rotation is using random starting index. I have no idea how we should test that the rotation is done at class level (rather than instance level as it was before)... I had a test like this, but it will fail because of the randomness:

```
    context 'with several objects' do
      let!(:response1){ described_class.new.get( '/test', body: '' ) }
      let!(:response2){ described_class.new.get( '/test', body: '' ) }
      let!(:response3){ described_class.new.get( '/test', body: '' ) }

      it 'works' do
        expect(response1).not_to eq( response2 )
        expect(response1).to eq( response3 )
        expect(response3).not_to eq( response2 )
      end
    end
```

lib/fortnox/api/environment_validation.rb
```diff
@@ -12,21 +12,27 @@ class CircularQueue
 
         def initialize *items
           @queue = [ *items ]
```
ehannes: @d-Pixie should `@queue` also be a class variable?

d-Pixie: No that would be weird. It's set in the constructor, instance level operation, but if it is class level the contents of that list could change when another instance is created and your counter could end up outside the new lists end ... We could argue that all of the state in that class should be class level, but how to you set the list then?

lib/fortnox/api/environment_validation.rb
```diff
@@ -12,21 +12,27 @@ class CircularQueue
...
-            @next_index += 1
+            @@next_index += 1
           end
           return value
         end
+
+        private
+
+          def random_start_index
+            Random.rand(0..@queue.size - 1)
```
d-Pixie: `Random.rand( @queue.size )` does the same ..

d-Pixie: `Random.rand( @queue.size )` does the same ..

PR #92, opened 2016-10-26
Implements gem exceptions

ehannes:


ehannes:
I have started to move some nested models from `fortnox/api/models/` to `fortnox/api/types`, but what should we do with models that are both nested and standalone models (for instance [`Label`](https://developer.fortnox.se/documentation/resources/labels/))?

d-Pixie:
If it has an endpoint in the API it's a model. Using nested models shouldn't pose a problem, it was only the row implementations that caused problems by not inheriting from our model base class.. 

So implement label as a proper model and subclass it from model base. That should work with the exception re-raising, no problem.

ehannes:
Now would be a good time to decide how we should test our models and types. The classes moved to the `Types` module was using custom model matchers and those specs are now broken.

lib/fortnox/api/models/base.rb
```diff
@@ -19,8 +19,12 @@ def self.attribute( name, *args )
         end
 
         def self.new( hash )
-          obj = preserve_meta_properties( hash ) do
-            super( hash )
+          begin
+            obj = preserve_meta_properties( hash ) do
+              super( hash )
+            end
+          rescue Dry::Struct::Error, Dry::Types::ConstraintError, Dry::Types::MissingKeyError => e
```
d-Pixie: We should catch these independently and preserve the information. That means adding two additional exception types and changing this to:

```
rescue Dry::Struct::Error => e
  raise Fortnox::API::AttributeError.new e
rescue Dry::Types::ConstraintError => e
  raise Fortnox::API::InvalidAttributeValueError.new e
rescue Dry::Types::MissingKeyError => e
  raise Fortnox::API::MissingAttributeError.new e
end
```

d-Pixie:
Other than that it LGTM ...

ehannes:
Rebased from `development` since I merged the another PR. Thats why the commit history looks a bit weird now.

ehannes:
@d-Pixie Two errors left, both concerning the exception handling. I think it's best if you could take a look.

ehannes:
Rubocop fixed the following with auto-correct: `Inherit from RuntimeError instead of Exception`. Hope that's ok.

ehannes:
I saw now that the exceptions were not inheriting from standard `Exception`, but from `Fortnox::API::Exception`. I stated it explicitly and Rubocop will not warn about this anymore.

ehannes:
@d-Pixie are there any more model tests we should do? Take a look at `fortnox/api/models/examples/model.rb`. Right now it just tests that a model can be initialized and tests the `unique_id` method. (Maybe that's all there is to test in these classes?)

lib/fortnox/api/models/base.rb
d-Pixie: Where they are, in model base, unless you know a reason for them to be elsewhere?

d-Pixie:
I think that is it for model testing for now Hannes :)

I'll have a look at the remaining errors.

lib/fortnox/api/models/base.rb
ehannes: We moved a lot of code to the `Types` module, and `Types::Model` raises `Fortnox::API::MissingAttributeError`, so I just wanted to make sure that this logic should be placed in `Model::Base`.

ehannes: If I implement the above, I get some errors which I think is related to the other failing specs. I'll push what I have and let you have a look at that too @d-Pixie :)

ehannes:
Travis build succeeded for all versions but Ruby 2.4.0-preview2 (for more information, see #93). We need to merge this as soon as possible and we cannot do anything about this right now, so lets leave the gem incompatible with version 2.4 and fix it later.

PR #94, opened 2016-11-10
Updates the readme to new codebase

ehannes:
- Adds new status as of November 2016.
- Includes notes about multiple Access Tokens.
- Validations are removed
- Mappers and Types are added


ehannes:
The information under Usage - Repositories is not correct. Maybe we should remove it for now and add it once we've implemented that?

> If you are eagle eyed you might have spotted the different classes for the entities returned in a collection vs the one we get from find. The Simple version of a class is used in thouse cases where the API-server doesn't return a full set of attributes for an entity. For customers the simple version has 10 attributes while the full have over 40.

PR #98, opened 2017-06-16
Fixes #97 Search with special character

### LINKED ISSUE #97: Search with special character crashes the gem
If you do a search with the gem, the parameter is not URI encoded and crashes the gem. A `Customer` search with name `däck` gives `URI::InvalidURIError (URI must be ascii only "/customers/?name=d\u{e4}ck")`

ehannes:


ehannes:
Hmm, all Travis builds for supported Ruby versions fail on the same three examples (the order differs since specs are run in random order):
```
Failures:
  1) Fortnox::API get access token
   ...
  2) Fortnox::API::Base making a request including the proper headers
  ...
  3) Fortnox::API::Base making requests with multiple access tokens with subsequent requests on same object works
  ...
```

These errors seem unrelated to the changes in this branch and all specs runs successfully on my computer...

PR #101, opened 2017-07-12
Test commit for debugging build failures with VCR

ehannes:


ehannes:
I can now reproduce the issue on my local machine. PR not needed. (This branch is only used for testing and will not be merged, at least not now).

PR #107, opened 2017-07-12
Project

### LINKED ISSUE #99: Add support for Fortnox Project

ehannes:
Fixes #99


ehannes:
I took the opportunity and rewrote the model-specs:en a bit. Most of the argument sent to the shared example group were no longer necessary since they were available as constants on the model classes themselves.

d-Pixie:
LGTM 👍

PR #108, opened 2017-07-24
Version 0.3.0

ehannes:


ehannes:
Wops, this PR has the wrong direction. I'll create a new one.

PR #111, opened 2017-08-09
Adds Exception info to README

ehannes:
Adds info about what Exceptions the gem can throw. Feel free to comment and improve this @d-Pixie :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ This is how all the models work, they are all immutable.
 
 ### Exceptions
 
-Models can throw `Fortnox::API::AttributeError` if an attribute is invalid in some way (for instance if you try to assigning an integer to a string attribute) and `Fortnox::API::MissingAttributeError` if a required attribute is missing.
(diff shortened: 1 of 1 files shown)


ehannes:
I'm not actually sure when `AttributeError` is thrown. This is the code:

```
# Fortnox::API::Model::Base
rescue Dry::Struct::Error => e
  raise Fortnox::API::AttributeError.new e
end
```

So the question is, when is `Dry::Struct::Error` thrown? The example in the README is not correct (assigning an integer to a string attribute does not throw any error, I think the invalid data is simply ignored).

d-Pixie:
Since we added a lot of types that do coercion that example might not work since the type now converts the integer to a string using `to_s`. Trying a similar assignment on a type that explicitly does not do coercion should throw the error though.
You can always look at the Dry::Types docs for examples of when the error is thrown :)

ehannes:
Actually, I found a case when `AttributeError` is thrown. If you try assigning 51 characters to an attribute that has a maximum of 50 characters.

d-Pixie:
It's fine for now. I'll give the entire README a once over before v1 and make sure we have examples for everything and so on. This is good enough for now at least :)

PR #113, opened 2017-10-08
Add Article resource

felixyz:
Pleasantly easy to add this resource along with a new enum type! :)

I was unable to use the `'.save'` example group because Fortnox adds `EUAccount`, `EUVATAccount`, `ExportAccount`, `PurchaseAccount` and `SalesAccount` values that don't correspond to actual accounts, thus `'old (update existing)'` fails. Any ideas?

### DIFF
--- a/lib/fortnox/api/models/article.rb
+++ b/lib/fortnox/api/models/article.rb
@@ -1,0 +1,133 @@
+require "fortnox/api/types"
+require "fortnox/api/models/base"
+
+module Fortnox
+  module API
+    module Model
+      class Article < Fortnox::API::Model::Base
+
+        UNIQUE_ID = :article_number
+        STUB = { description: '' }.freeze
+
+        #Url Direct URL to the record
+        attribute :url, Fortnox::API::Types::Nullable::String.with( read_only: true )
+
+        #Active If the article is active
+        attribute :active, Fortnox::API::Types::Nullable::Boolean
+
+        #ArticleNumber Article number. 50 characters
+        attribute :article_number, Fortnox::API::Types::Sized::String[ 50 ]
+
+        #Bulky If the article is bulky.
+        attribute :bulky, Fortnox::API::Types::Nullable::Boolean
+
+        #ConstructionAccount Account number for construction work (special VAT rules in Sweden).
+        # The number must be of an existing account.
+        attribute :construction_account, Types::Sized::Integer[ 0, 9_999 ]
+
+        #Depth The depth of the article in millimeters
+        attribute :depth, Types::Sized::Integer[ 0, 99_999_999 ]
+
+        #Description The description of the article
+        attribute :description, Fortnox::API::Types::Sized::String[ 200 ].with( required: true )
+
+        #DisposableQuantity Disposable quantity of the article.
+        attribute :disposable_quantity, Fortnox::API::Types::Nullable::Float.with( read_only: true )
+
+        #EAN EAN bar code
+        attribute :ean, Fortnox::API::Types::Sized::String[ 30 ]
+
+        #EUAccount Account number for the sales account to EU.
+        # The number must be of an existing account.
+        attribute :eu_account, Types::Sized::Integer[ 0, 9_999 ]
+
+        #EUVATAccount Account number for the sales account to EU with VAT.
+        # The number must be of an existing account.
+        attribute :eu_vat_account, Types::Sized::Integer[ 0, 9_999 ]
+
+        #ExportAccount Account number for the sales account outside EU
+        # The number must be of an existing account.
+        attribute :export_account, Types::Sized::Integer[ 0, 9_999 ]
+
+        #Height The height of the article in millimeters
+        attribute :height, Types::Sized::Integer[ 0, 99_999_999 ]
+
+        #Housework If the article is housework
+        attribute :housework, Fortnox::API::Types::Nullable::Boolean
+
+        #HouseWorkType The type of house work.
+        attribute :house_work_type, Types::HouseWorkType
+
+        #Manufacturer The manufacturer of the article
+        attribute :manufacturer, Fortnox::API::Types::Sized::String[ 50 ]
+
+        #ManufacturerArticleNumber The manufacturer's article number
+        attribute :manufacturer_article_number, Fortnox::API::Types::Sized::String[ 50 ]
+
+        #Note Text note
+        attribute :note, Fortnox::API::Types::Sized::String[ 10_000 ]
+
+        #PurchaseAccount Account number for purchase.
+        # The number must be of an existing account.
+        attribute :purchase_account, Types::Sized::Integer[ 0, 9_999 ]
+
+        #PurchasePrice Purchase price of the article
+        attribute :purchase_price, Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999_999.9 ]
+
+        #QuantityInStock Quantity in stock of the article
+        attribute :quantity_in_stock, Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999_999.9 ]
+
+        #ReservedQuantity Reserved quantity of the article
+        attribute :reserved_quantity, Fortnox::API::Types::Nullable::Float.with( read_only: true )
+
+        #SalesAccount Account number for the sales account in Sweden.
+        # The number must be of an existing account.
+        attribute :sales_account, Types::Sized::Integer[ 0, 9_999 ]
+
+        #SalesPrice Price of article for its default price list
+        attribute :sales_price, Fortnox::API::Types::Nullable::Float.with( read_only: true )
+
+        #StockGoods If the article is stock goods
+        attribute :stock_goods, Fortnox::API::Types::Nullable::Boolean
+
+        #StockPlace Storage place for the article
+        attribute :stock_place, Fortnox::API::Types::Sized::String[ 100 ]
+
+        #StockValue Value in stock of the article
+        attribute :stock_value, Fortnox::API::Types::Nullable::Float.with( read_only: true )
+
+        #StockWarning When to start warning for low quantity in stock
+        attribute :stock_warning, Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999_999.9 ]
+
+        #SupplierName Name of the supplier
+        attribute :supplier_name, Fortnox::API::Types::Nullable::String.with( read_only: true )
+
+        #SupplierNumber Supplier number for the article.
+        # The number must be of an existing supplier.
+        attribute :supplier_number, Fortnox::API::Types::Nullable::String
+
+        #Type The type of the article
+        attribute :type, Types::ArticleType
+
+        #Unit Unit code for the article.
+        # The code must be of an existing unit.
+        attribute :unit, Fortnox::API::Types::Nullable::String
+
+        #VAT VAT percent, this is predefined by the VAT for the sales account
+        attribute :vat, Fortnox::API::Types::Nullable::Float
+
+        #WebshopArticle If the article is a webshop article
+        attribute :webshop_article, Fortnox::API::Types::Nullable::Boolean
+
+        #Weight Weight of the article in grams
+        attribute :weight, Types::Sized::Integer[ 0, 99_999_999 ]
+
+        #Width Width of the article in millimeters.
+        attribute :width, Types::Sized::Integer[ 0, 99_999_999 ]
+
+        #Expired If the article has expired
+        attribute :expired, Fortnox::API::Types::Nullable::Boolean
+      end
+    end
+  end
+end
--- a/spec/fortnox/api/mappers/article_spec.rb
+++ b/spec/fortnox/api/mappers/article_spec.rb
@@ -1,0 +1,15 @@
+require 'spec_helper'
+require 'fortnox/api'
+require 'fortnox/api/mappers/customer'
+require 'fortnox/api/mappers/examples/mapper'
+
+describe Fortnox::API::Mapper::Article do
+  key_map = Fortnox::API::Mapper::Article::KEY_MAP
+
+  json_entity_type = 'Article'
+  json_entity_collection = 'Articles'
+
+  it_behaves_like 'mapper', key_map, json_entity_type, json_entity_collection do
+    let(:mapper){ described_class.new }
+  end
+end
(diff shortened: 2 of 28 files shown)


lib/fortnox/api/types/model.rb
```diff
@@ -24,14 +24,10 @@ def missing_keys( attributes )
         end
 
         def first_missing_required_key( attributes )
```
felixyz: Unrelated stylistic change. Could be removed from the PR if that's preferable?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+        #     :supplier_number
+        # )
+        #
+        # sortable_attributes(
+        #   :article_number, :disposable_quantity, :reserved_quantity, :stock_value,
+        #   :quantity_in_stock
+        # )
+        #
+        # filter(
+        #   :active
```
felixyz: Not sure why models don't list `:inactive` as well?

spec/fortnox/api/repositories/article_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  include_examples '.save with specially named attribute',
+                   { description: 'Test article' },
+                   :ean,
+                   '5901234123457'
+
+  include_examples '.all', 12
+
+  include_examples '.find', '1' do
+    let( :find_by_hash_failure ){ { description: 'Not Found' } }
+    let( :single_param_find_by_hash ){ { find_hash: { articlenumber: 1 }, matches: 3 } }
```
felixyz: Took me a while to figure out it should be `articlenumber` rather than `article_number`. Wouldn't it be preferable to use the same mappings as for the model itself?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #DisposableQuantity Disposable quantity of the article.
+        attribute :disposable_quantity, Fortnox::API::Types::Nullable::Float.with( read_only: true )
+
+        #EAN EAN bar code
+        attribute :ean, Fortnox::API::Types::Sized::String[ 30 ]
+
+        #EUAccount Account number for the sales account to EU.
+        # The number must be of an existing account.
+        attribute :eu_account, Types::Sized::Integer[ 0, 99_99 ]
```
ehannes: Would you mind changing this to `9_999` to keep it consistent?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #EUVATAccount Account number for the sales account to EU with VAT.
+        # The number must be of an existing account.
+        attribute :eu_vat_account, Types::Sized::Integer[ 0, 99_99 ]
```
ehannes: Would you mind changing this to `9_999` to keep it consistent?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #ExportAccount Account number for the sales account outside EU
+        # The number must be of an existing account.
+        attribute :export_account, Types::Sized::Integer[ 0, 99_99 ]
```
ehannes: Would you mind changing this to `9_999` to keep it consistent?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #ManufacturerArticleNumber The manufacturer's article number
+        attribute :manufacturer_article_number, Fortnox::API::Types::Sized::String[ 50 ]
+
+        #Note Text note
+        attribute :note, Fortnox::API::Types::Sized::String[ 10_000 ]
+
+        #PurchaseAccount Account number for purchase.
+        # The number must be of an existing account.
+        attribute :purchase_account, Types::Sized::Integer[ 0, 99_99 ]
```
ehannes: Would you mind changing this to `9_999` to keep it consistent?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #PurchasePrice Purchase price of the article
+        attribute :purchase_price, Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999_999.9 ]
```
ehannes: @d-Pixie & @felixyz Accoding to the documentation, this attribute is a `float, 14 digits`. Should the decimal fraction be `.9` as here, or should it be `.0` as in `Fortnox::API::Types::DocumentRow`: `attribute :delivered_quantity, Types::Sized::Float[ 0.0, 9_999_999_999_999.0 ]`?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #QuantityInStock Quantity in stock of the article
+        attribute :quantity_in_stock, Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999_999.9 ]
+
+        #ReservedQuantity Reserved quantity of the article
+        attribute :reserved_quantity, Fortnox::API::Types::Nullable::Float.with( read_only: true )
+
+        #SalesAccount Account number for the sales account in Sweden.
+        # The number must be of an existing account.
+        attribute :sales_account, Types::Sized::Integer[ 0, 99_99 ]
```
ehannes: Would you mind changing this to `9_999` to keep it consistent?

lib/fortnox/api/models/article.rb
```diff
@@ -0,0 +1,147 @@
...
+
+        #StockWarning When to start warning for low quantity in stock
+        attribute :stock_warning, Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999_999.9 ]
+
+        #SupplierName Name of the supplier
+        attribute :supplier_name, Fortnox::API::Types::Nullable::String.with( read_only: true )
+
+        #SupplierNumber Supplier number for the article.
+        # The number must be of an existing supplier.
+        attribute :supplier_number, Fortnox::API::Types::Nullable::String.with( read_only: true )
```
ehannes: Should not be read only according to the documentation.

ehannes:
About the `save`-example group: I could take a look at it. The specs for the repositories do need some improvements...

spec/fortnox/api/repositories/article_spec.rb
felixyz:
```Shell
[1] pry(main)> Fortnox::API::Repository::Article.new.find(article_number: 1)
Fortnox::API::RemoteServerError: Ogiltig parameter i anropet.
from /Users/felix/code/fortnox-api/lib/fortnox/api/request_handling.rb:12:in `raise_api_error'
[2] pry(main)> Fortnox::API::Repository::Article.new.find(articlenumber: 1)
=> [#<Fortnox::API::Model::Article url=nil active=nil article_number="0001" bulky=nil
```

ehannes: Thanks. Something is wrong here... You should be able to use `article_number`.

ehannes: What does the `.yml`-cassette look like?

lib/fortnox/api/models/article.rb
felixyz: I got this from the declaration of `InvoiceAdministrationFee` (float, 11 digits) in Customer:

```Ruby
attribute :invoice_administration_fee,
  Fortnox::API::Types::Sized::Float[ 0.0, 99_999_999_999.9 ]
```

(11 + 1)

But I really have no idea what the correct definition would be. Seems to me the number of _digits_ in a float can't be captured by a number range?

lib/fortnox/api/models/article.rb
felixyz: Too much copy-paste 😁

spec/fortnox/api/mappers/article_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+require 'fortnox/api'
+require 'fortnox/api/mappers/customer'
+require 'fortnox/api/mappers/examples/mapper'
+
+describe Fortnox::API::Mapper::Article do
+  key_map =
```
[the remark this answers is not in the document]
felixyz: Same goes for the `KEY_MAP` in `Mapper::Article` right? Code Climate is complaining rather sensibly about this duplication. Do you want it to be explicit or have you considered DRYing it up?

spec/fortnox/api/repositories/article_spec.rb
felixyz: https://github.com/my-codeworks/fortnox-api/pull/113/files#diff-15da9a4e8da28c13ef0e0a4b2e8e85a6R1

felixyz: The exact same thing happens with the customer repo:

```
> Fortnox::API::Repository::Customer.new.find(organisation_number: 1)
Fortnox::API::RemoteServerError: Ogiltig parameter i anropet.
```

Works with `customernumber`.

spec/fortnox/api/mappers/article_spec.rb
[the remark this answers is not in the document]
d-Pixie: If all you need is the actual key map from the class you can use that directly, or clone it, as: `Fortnox::API::Mapper::Article::KEY_MAP`.

spec/fortnox/api/repositories/article_spec.rb
felixyz: (Personally I would probably like if it was very forgiving so that `customernumber`, `customer_number`, `CustomerNumber` produced the same request. But if you want it more restrictive `customer_number` seems like the obvious choice.)

lib/fortnox/api/models/article.rb
d-Pixie: It should not really be in the codebase. It was added when we worked on the filtering feature. But commented out code should not be in the repo, ever. So this should be removed. Sorry for setting a bad example there @felixyz :)

spec/fortnox/api/repositories/article_spec.rb
d-Pixie: In an effort to keep client code consistent and in keeping with idiomatic Ruby I think we'll only allow for one variant, the underscored one :) But that one should work, so that is a bug...

lib/fortnox/api/models/article.rb
felixyz: I personally thought it was pretty helpful to have documentation for searchable/sortable but I guess the official Fortnox docs are a better source anyway.

felixyz:
@ehannes Are the `save` specs supposed to delete the newly created resource in a clean-up step? (For those resource types that support it.) I found some notes that seemed to suggest that but after a quick look found nothing, so my spec trials left a few extra articles in the system.

spec/fortnox/api/mappers/article_spec.rb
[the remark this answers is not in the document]
felixyz: I went with `key_map = Fortnox::API::Mapper::Article::KEY_MAP` which seems adequate in my (limited) understanding.

ehannes:
@felixyz We have a slack channel. Would you be interested in joining us?

felixyz:
@ehannes Think I fixed all your feedback. `.rubocop.yml` is no longer compatible with latest `rubocop 0.4x` (which is 0.49.1) so needs to be pinned down in gemspec or (better) fix `.rubocop.yml`.

felixyz:
@ehannes Slack – yes sure!

ehannes:
Should I invite you on your github mail?

felixyz:
Yes please

d-Pixie:
@felixyz You have gotten an invite.

ehannes:
> Are the save specs supposed to delete the newly created resource in a clean-up step? (For those resource types that support it.) I found some notes that seemed to suggest that but after a quick look found nothing, so my spec trials left a few extra articles in the system.

Good catch. The `.save`-spec currenly only adds resources, so there is no clean-up step, since not all resources supports deleting. It would certainly be better if our specs had a clean-up step :)

ehannes:
> @ehannes Think I fixed all your feedback. .rubocop.yml is no longer compatible with latest rubocop 0.4x (which is 0.49.1) so needs to be pinned down in gemspec or (better) fix .rubocop.yml.

Thanks for mentioning it! I'll look into it.

ehannes:
TODO:s

- [ ] .rubocop.yml is no longer compatible with latest rubocop 0.4x (which is 0.49.1) so needs to be pinned down in gemspec or (better) fix .rubocop.yml.
- [ ] Shared example group `.save`.
- [ ] Attributes with floating point limits.
- [ ] Repository find and `articlenumber`.

spec/fortnox/api/repositories/article_spec.rb
ehannes: @felixyz & @d-Pixie 
Ok, now I know how this works.

First of all, Fortnox expects the request to look something like this:
```
https://api.fortnox.se/3/articles?articlenumber=1
```

Now to the gem. The `find` method calls `find_all_by` when you send it a hash.
```
        def find_all_by( hash )
          response_hash = get( "#{ self.class::URI }?#{ to_query( hash ) }" )
          instantiate_collection_response( response_hash )
        end
```

There are no use of any mappers or anything here (`to_query` simply turns `articlenumber: 1` to `&articlenumber=1`). Since `find_all_by` simply turns whatever you give it to query parameters, it not only supports attributes, but filters, sorting, searching, limits & offsets as well as pagination. That is why we need to pass `articlenumber` and not `article_number` to the `find`-method. Whether we should change that or not is open for discussion :)

ehannes:
@felixyz About the `.save` example group. I get this error message:
```
  1) Fortnox::API::Repository::Article.save old (update existing) includes correct :description
     Failure/Error: raise Fortnox::API::RemoteServerError, message
     
     Fortnox::API::RemoteServerError:
       Försäljningskonto inte uppdaterat. Kontot "3011" existerar inte. (SalesAccount)
```

Since we have not set the `SalesAccount` for this `Article`, I guess Fortnox tries to set the default SalesAccount. Apparently, that is 3011, and that does not exist. To get this test working, we need to login to our Fortnox Sandbox account and add this 3011-account. I'll try to do it and see if that helps :)

spec/fortnox/api/repositories/article_spec.rb
ehannes: @felixyz We'll try to fix this in the future, but for now we need to use `articlenumber`. Thanks for the notice!

ehannes: See #117

ehannes:
About the floating points, see Floating point limits #116.

PR #120, opened 2017-10-12
Command for quick CLI testing during development

felixyz:
Allows you to get a Pry session with everything you need to test the gem with

`$ bin/console`


d-Pixie:
I really like this one.

PR #123, opened 2017-11-16
WiP - Support multiple token stores

d-Pixie:
Adds basic support tu use several token stores.


d-Pixie:
@ehannes is opening a new PR that uses http://dry-rb.org/gems/dry-configurable/ instead of my homebrew solution 😄

PR #124, opened 2017-12-06
Multi token

ehannes:
Introduces support for multiple Fortnox backends. Each Fortnox Access Token is connected to a specific Fortnox instance.You can now call different Fortnox backends with the same gem instance. For instance (where `fortnox1` and `fortnox2` represents key sets for different Fortnox instances):

```ruby
  config.client_secret = 'your-secret'
  config.token_store = { fortnox1: ['token1', 'token2'], fortnox2: ['token1', 'token2'] }
end

Fortnox::API::Repository::Customer.new( token_store: :fortnox_instance_1 )
```

This pull request changes a lot of stuff, both internally and public:
- Removes support for fetching Access Tokens with the gem (we can easily implement it again if we want, but for now let's leave it outside the gem until someone explicitly asks for it. No need to include features that no one uses).
- The gem is no longer configured with environment variables, but with a configuration
- Dependency to the Dotenv gem is removed.
- `Fortnox::API::Base` is merged with `Fortnox::API::Repository::Base`, the former is no longer needed.

TODO:
- [x] Update Readme
- [x] Fix support for default token store


ehannes:
I have not yet found a solution for how to default to `:default` token store when no explicit `token_store` is configured:

This works:
```
  config.token_store = { default: ['token1', 'token2'], store2: ['token1', 'token2'] }
end

Fortnox::API::Repository::Customer.new # Uses token store `:default`
Fortnox::API::Repository::Customer.new( token_store: :store2 ) # Uses token store `:store2`
```

I have no solution for this yet:
```
  config.access_token = ['token1', 'token2']
end

This should set `Fortnox::API.config.token_store` to `{ default: ['token1', 'token2'] }`, but I have not found any working implementation for it yet.
```

ehannes:
@d-Pixie ready for code review :)

README.md
```diff
@@ -114,44 +114,46 @@ $ gem install fortnox-api
...
-require 'fortnox/api/access_token'
-
-Fortnox::API::AccessToken.new(
-  base_url: 'https://api.fortnox.se/3',
-  client_secret: 'P5K5wE3Kun', # Replace with your client secret
-  authorization_code: 'ea3862b1-189c-464b-8e25-1b9702365fa1', # Replace with your auth code
-)
+Fortnox::API.configure do |config|
+  config.client_secret = 'P5K5wE3Kun'
+  config.access_tokens = ['3f08d038-f380-4893-94a0a0-8f6e60e67a']
```
d-Pixie: Can we get this to be `config.access_token = '3f08d038-f380-4893-94a0a0-8f6e60e67a'` and just alias it over to `config.access_tokens` in the setter? Not sure how much dry config will let you do, but it would be a lot nicer.

README.md
```diff
@@ -114,44 +114,46 @@ $ gem install fortnox-api
...
 
-### Multiple AccessTokens
-
-Fortnox uses quite low [API rate limits](https://developer.fortnox.se/blog/important-implementation-of-rate-limits/). The limit is for each access token, and according to Fortnox you can use as many tokens as you like to get around this problem. This gem supports multiple access tokens automatically. Just separate them with a comma: `FORTNOX_API_ACCESS_TOKEN=a78d35hc-j5b1-ga1b-a1h6-h72n74fj5327,s2b45f67-dh5d-3g5s-2dj5-dku6gn26sh62` and the gem will automatically rotate between these tokens. In theory you can declare as many as you like. Remember that you will need one authorization code for each token! See Getting an AccessToken above.
+```ruby
+Fortnox::API.configure do |config|
+  config.client_secret = 'P5K5wE3Kun'
+  conig.access_tokens = {
```
d-Pixie: `conig`? :)

lib/fortnox/api/mappers/base/from_json.rb
```diff
@@ -45,7 +45,7 @@ def convert_collection( key, collection )
             # add it before a new release.
 
             raise MissingModelOrMapperException, "for #{ key } (#{ mapper_name }, #{ Fortnox::API::Mapper::DefaultTemplates.canonical_name_sym }) with #{ collection }" if ENV['RUBY_ENV']
-            Fortnox::API.logger.warn( "Missing Model or Mapper implementation for #{ key } with attributes: #{ collection }" )
+            Fortnox::API.config.logger.warn( "Missing Model or Mapper implementation for #{ key } with attributes: #{ collection }" )
```
d-Pixie: You could have aliased this on the root module instead (made logger there a class method that looks up the logger in the config).

lib/fortnox/api/repositories/base.rb
```diff
@@ -1,22 +1,70 @@
...
+
+        def check_access_tokens!( tokens )
+          if tokens == nil or tokens.length.zero?
+            fail MissingConfiguration, "You have not provided any access tokens in token store #{ @token_store.inspect }."
+          end
+        end
+
+        def get_access_tokens
+          raw_tokens = config.access_tokens
+          tokens = raw_tokens.is_a?(Hash) ? get_hash_access_tokens(raw_tokens) : raw_tokens
```
d-Pixie: It would be cleaner if the config took care of this for us, see the comment above for a single access token.

We could define three setters for the config: `access_token`, `access_tokens` and `token_store` where the first takes a string, the second takes an array and the third a hash. The first two would then in actuality insert the config value into the `token_store` config value by appropriately wrapping the value as `{ default: [ value ]}` and `{ default: value }` respectively.

That way we only ever need to handle one case in our code.

lib/fortnox/api/repositories/customer.rb
```diff
@@ -5,11 +5,14 @@
 module Fortnox
   module API
     module Repository
-      class Customer < Fortnox::API::Repository::Base
-
-        MODEL = Fortnox::API::Model::Customer
-        MAPPER = Fortnox::API::Mapper::Customer
+      class Customer < Repository::Base
```
d-Pixie: If you are cutting specificity this could just be `Base`?

lib/fortnox/api/request_handling.rb
```diff
@@ -7,7 +7,7 @@ module RequestHandling
         def raise_api_error( error, response )
           message = ( error[ 'message' ] || error[ 'Message' ] || 'Okänt fel' )
 
-          message += "\n\n#{ response.request.inspect }" if Fortnox::API.debugging
+          message += "\n\n#{ response.request.inspect }" if Fortnox::API.config.debugging
```
d-Pixie: This could also be aliased, like `logger`.

spec/fortnox/api/repositories/base_spec.rb
```diff
@@ -0,0 +1,329 @@
...
+  using_test_class do
+    module Model
+      class Test
+      end
+    end
+  end
+
+  # Note: Could not get this to work with using_test_classes
+  # I get Dry::Container::Error:
+  # There is already an item registered with the key "test"
```
d-Pixie: You need to clean the container manually after each test.

README.md
ehannes: I'll have a look. I had that in mind too, but I wanted you to take a look at the code first :)

lib/fortnox/api/repositories/base.rb
ehannes: That would be nice, but where to put this code? This cannot be done in `Fortnox::API` since it's a module. Wouldn't we need to do this logic in `Fortnox::API::Repository::Base#initialize`?

d-Pixie: I would like that to be supported by the configuration module, but I'm not sure it is :) You will have to play with the processor version of declaring a setting:
```
setting :token_store, {}

setting(:access_token) do |value|
  # some hack to set @_config['token_store'] to { default: [ value ]}
end

setting(:access_tokens) do |value|
  # some hack to set @_config['token_store'] to { default: value }
end
`
``
But as I said, I'm not sure it's possible.

Comment (ehannes) [reply lib/fortnox/api/repositories/base.rb]: I did try something like this, but that did not work.
```ruby
setting :token_store, { default: self.config.access_token }
```
I've also looked at the [gem documentation](http://dry-rb.org/gems/dry-configurable/) without finding anything about it. I'll continue to play around and look for a hack :)

ehannes:
@d-Pixie What do you say about the methods in `Fortnox::API::Repository::Base`? I would argue that most of them should be private/protected since they are not meant to be used from outside.

lib/fortnox/api/repositories/base.rb
ehannes: I think I've found a solution for this now :) Implementation on the way.

ehannes:
@d-Pixie Time to take another look at this :)

What do you say about the methods in `Repository::Base`? I would argue that most of them should be private/protected.

spec/fortnox/api/repositories/base_spec.rb
```diff
@@ -0,0 +1,323 @@
...
+        subject{ access_tokens }
+        before{ Fortnox::API.configure{ |conf| conf.access_tokens = access_tokens } }
+        let( :access_tokens ){ [access_token, access_token2] }
+
+        it{ is_expected.to include( repository.next_access_token ) }
+
+        # rubocop:disable RSpec/MultipleExpectations
+        # All these checks must be run in same it-statement because
+        # of the random starting index.
+        it 'circulates the tokens' do
```
d-Pixie: Break into two tests.

lib/fortnox/api/repositories/base.rb
```diff
@@ -1,22 +1,74 @@
...
+          @mapper = Registry[ Mapper::Base.canonical_name_sym( model )].new
+        end
+
+        def next_access_token
+          @access_tokens ||= CircularQueue.new( *get_access_tokens )
+          @access_tokens.next
+        end
+
+        def check_access_tokens!( tokens )
+          if tokens == nil or tokens.length.zero?
```
d-Pixie: `tokens.empty?`

lib/fortnox/api/repositories/base.rb
```diff
@@ -1,22 +1,74 @@
...
+            fail MissingConfiguration, "You have not provided any access tokens in token store #{ @token_store.inspect }."
+          end
+        end
+
+        def get_access_tokens
+          begin
+            tokens = config.token_store.fetch( @token_store )
+          rescue KeyError
+            no_token_store_found!
```
d-Pixie: Not really true, THAT token store was not found, but there might be others ...

PR #131, opened 2018-02-16
Updates rubocop version and removes local config

d-Pixie:


.rubocop.yml
```diff
@@ -1,1116 +0,0 @@
-require: rubocop-rspec
```
ehannes: We want to keep this.

.rubocop.yml
```diff
@@ -1,1116 +0,0 @@
...
-
-AllCops:
```
ehannes: And this.

.rubocop.yml
```diff
@@ -1,1116 +0,0 @@
...
-Style/WordArray:
-  Description: 'Use %w or %W for arrays of words.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#percent-w'
-  Enabled: false
-
-#################### RSpec ################################
-RSpec/FilePath:
-  Enabled: false
-
-RSpec/NestedGroups:
```
ehannes: We should keep this as well.

.rubocop.yml
```diff
@@ -1,1116 +0,0 @@
...
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#while-as-a-modifier'
-  Enabled: false
-
-Style/WordArray:
-  Description: 'Use %w or %W for arrays of words.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#percent-w'
-  Enabled: false
-
-#################### RSpec ################################
-RSpec/FilePath:
```
ehannes: I'm not sure we added this. We'll need to try removing it and see what happens :)

Issue #32, opened 2016-03-09
Attribute 'url' is nil when fetching Customer


ehannes:
When requesting all `Customer`s from Fortnox through the gem, this is the first entry:

```
{"url":null,"address1":"","address2":"","city":"","country":null,"comments":null,"currency":null,"cost_center":null,"country_code":null,"customer_number":"1","delivery_address1":null,"delivery_address2":null,"delivery_city":null,"delivery_country":null,"delivery_country_code":null,"delivery_fax":null,"delivery_name":null,"delivery_phone1":null,"delivery_phone2":null,"delivery_zip_code":null,"email":"","email_invoice":null,"email_invoice_bcc":null,"email_invoice_cc":null,"email_offer":null,"email_offer_bcc":null,"email_offer_cc":null,"email_order":null,"email_order_bcc":null,"email_order_cc":null,"fax":null,"invoice_administration_fee":null,"invoice_discount":null,"invoice_freight":null,"invoice_remark":null,"name":"Test","organisation_number":"","our_reference":null,"phone1":null,"phone2":null,"price_list":null,"project":null,"sales_account":null,"show_price_vat_included":null,"terms_of_delivery":null,"terms_of_payment":null,"type":null,"vat_number":null,"vat_type":null,"visiting_address":null,"visiting_city":null,"visiting_country":null,"visiting_zip_code":null,"way_of_delivery":null,"your_reference":null,"zip_code":""}
```

When I do the same request with `curl`, this is the first entry:

```
{"@url":"https:\/\/api.fortnox.se\/3\/customers\/1","Address1":"","Address2":"","City":"","CustomerNumber":"1","Email":"","Name":"Test","OrganisationNumber":"","Phone":"","ZipCode":""}
```

Many attributes are filtered out when doing an `all`-request to Fortnox. Our gem sets these attributes to `null` instead, which can be incorrect but I think #25 is meant to fix that.

Instead, this issue concerns the `url` attribute. Gem sets it to `null`, but Fortnox returns `https:\/\/api.fortnox.se\/3\/customers\/1`. This is obviously wrong, but should our models even have a `url` attribute? That attribute declares where the resource can be found at Fortnox, but when using this gem we are not interested in querying directly against Fortnox. Or should we have some find method that takes the `url` as an argument and fetches that resource?

ehannes:
See #50.

ehannes:
[...]

PR #133, opened 2018-02-22
Adds unit

d-Pixie:


lib/fortnox/api/mappers/unit.rb
```diff
@@ -0,0 +1,15 @@
...
+module Fortnox
+  module API
+    module Mapper
+      class Unit < Fortnox::API::Mapper::Base
+        KEY_MAP = {}.freeze
+        JSON_ENTITY_WRAPPER = 'Unit'.freeze
+        JSON_COLLECTION_WRAPPER = 'Units'.freeze
+      end
+
+      Registry.register( Unit.canonical_name_sym, Unit )
```
ehannes: Maybe we should move the registry registration to `Fortnox::API::Mapper::Base` to avoid repeating this code in every mapper class?

lib/fortnox/api/models/unit.rb
```diff
@@ -0,0 +1,22 @@
...
+        STUB = { code: '' }.freeze
+
+        # Url Direct URL to the record
+        attribute :url, Types::Nullable::String.with(read_only: true)
+
+        # Comments Comments on project. 512 characters
+        attribute :code, Types::Nullable::String.with(read_only: true)
+
+        # ContactPerson ContactPerson for project. 50 characters
+        attribute :description, Types::Nullable::String
```
ehannes: Shouldn't the `50 characters`-limit be included?

lib/fortnox/api/models/unit.rb
```diff
@@ -0,0 +1,22 @@
...
+    module Model
+      class Unit < Fortnox::API::Model::Base
+        UNIQUE_ID = :code
+        STUB = { code: '' }.freeze
+
+        # Url Direct URL to the record
+        attribute :url, Types::Nullable::String.with(read_only: true)
+
+        # Comments Comments on project. 512 characters
+        attribute :code, Types::Nullable::String.with(read_only: true)
```
ehannes: Shouldn't the `512 characters`-limit be included?

lib/fortnox/api/models/unit.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module Fortnox
+  module API
+    module Model
+      class Unit < Fortnox::API::Model::Base
+        UNIQUE_ID = :code
+        STUB = { code: '' }.freeze
+
+        # Url Direct URL to the record
+        attribute :url, Types::Nullable::String.with(read_only: true)
```
ehannes: As mentioned in #32, this attribute is actually called `@url`, but I guess we postpone that fix to when we fix #32.

spec/fortnox/api/mappers/unit_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'fortnox/api/mappers/examples/mapper'
+
+describe Fortnox::API::Mapper::Unit do
+  key_map = Fortnox::API::Mapper::Unit::KEY_MAP
+
+  json_entity_type = 'Unit'
+  json_entity_collection = 'Units'
+
+  it_behaves_like 'mapper', key_map, json_entity_type, json_entity_collection do
+    let(:mapper){ described_class.new }
```
ehannes: Hmm, this argument is really meaningless. But that's for a later PR...

spec/fortnox/api/repositories/unit_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+require 'fortnox/api'
+require 'fortnox/api/mappers'
+require 'fortnox/api/repositories/customer'
+require 'fortnox/api/repositories/examples/all'
+require 'fortnox/api/repositories/examples/find'
```
ehannes: This import is currently not used. I guess that's because there are no searchable attributes in this model. Since we have a find method on this repository even though there are no searchable attributes, we should probably test that the `find`-method behaves as expected (probably by throwing some error).

lib/fortnox/api/mappers/unit.rb
d-Pixie: Maybe ... But that makes it more magical :)

lib/fortnox/api/models/unit.rb
d-Pixie: Yes.

spec/fortnox/api/mappers/unit_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+require 'fortnox/api'
+require 'fortnox/api/mappers/unit'
+require 'fortnox/api/models/unit'
+require 'fortnox/api/mappers/examples/mapper'
+
+module Fortnox
```
ehannes: Why the modules? The usual way to do this is to write `describe Fortnox::API::Mapper::Unit`.

spec/fortnox/api/mappers/unit_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  module API
+    # Shhh Rubocop, we don't need a comment here ... Really
```
ehannes: We'll fix this when we merge the Rubocop update. We should add a `TODO:` here

spec/fortnox/api/mappers/unit_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    module Mapper
+      describe Unit do
+        describe 'when mapping model' do
+          let(:model) { Model::Unit.new(code: 'lbs', description: 'Pounds') }
```
ehannes: I would change `Model::Unit` to `described_class::Model::Unit` and remove the module declarations in this spec.

spec/fortnox/api/mappers/unit_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+require 'fortnox/api/mappers/unit'
+require 'fortnox/api/models/unit'
+require 'fortnox/api/mappers/examples/mapper'
+
+module Fortnox
+  module API
+    # Shhh Rubocop, we don't need a comment here ... Really
+    module Mapper
+      describe Unit do
+        describe 'when mapping model' do
```
ehannes: I would use `context` here instead of `describe`. The convention is to use `describe` for describing methods, classes and such and use `context` to give a context to the test.

spec/fortnox/api/repositories/unit_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+require 'fortnox/api'
+require 'fortnox/api/mappers'
+require 'fortnox/api/repositories/unit'
+require 'fortnox/api/repositories/examples/all'
+require 'fortnox/api/repositories/examples/save'
+require 'fortnox/api/repositories/examples/save_with_specially_named_attribute'
+
```
ehannes: I would add a `TODO` in this spec to remember to test the `find` method here later on when we have a test for that.

spec/fortnox/api/models/unit_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+require 'fortnox/api/models/unit'
+
+module Fortnox
+  module API
+    # Shhh Rubocop, we don't need a comment here ... Really
```
ehannes: We'll fix this when we merge the Rubocop update. We should add a TODO: here

PR #139, opened 2018-02-24
Adds resource TermsOfPayments

ehannes:
This resource added some edge cases... I've added the following TODO:s to the model class:

```
# TODO: Only writable during POST.
# TODO: Must be alphanumeric
# TODO: 30days is a valid value, but the API rewrites it as 30DAYS
#       and you will not find the resource with a GET with value '30days'
```


ehannes:
I've added the two last TODO:s in our wiki since the information is missing from Fortnox documentation.

PR #144, opened 2018-03-01
Renames house work to housework.

d-Pixie:


d-Pixie:
@felixyz Thi change is breaking if you use the housework constants directly in your code. Just so you don't miss it :)

felixyz:
@d-Pixie I don't but thanks for the heads-up!

Issue #95, opened 2016-11-14
Correct attribute ranges


ehannes:
We need to go through our attributes and check the ranges. For instance, a Fortnox `Order` can have an `OrderRow` with negative `Price`, but our implementation does not support that:

`attribute :price, Types::Sized::Float[ 0.0, 99_999_999_999.0 ]`

ehannes:
Resolved long time ago 😄

PR #148, opened 2018-03-05
Allow delivered_quantity to be negative

mbirman:
Hi,

Many thanks for the gem, it's really useful.

I updated `delivered_quantity` attribute of the `DocumentRow` so it can be negative as well as over 9000.

Here is a real world example of why this is needed. When working with Systembolaget, we sometimes receive claims on goods that were sent by mistake, damaged during shipment etc. We have to include the claim sum on the next invoice, it looks like an invoice row with negative quantity. Fortnox allows it from the web version, and I believe it shouldn't be prohibited by the API wrapper.

Please review the changes and let me know if you have any questions or objections.

Cheers 🖖

### DIFF
--- a/lib/fortnox/api/types/document_row.rb
+++ b/lib/fortnox/api/types/document_row.rb
@@ -20,7 +20,7 @@ module Fortnox
         attribute :cost_center, Types::Nullable::String
 
         # DeliveredQuantity Delivered quantity. 14 digits
-        attribute :delivered_quantity, Types::Sized::Float[0.0, 9_999_999_999_999.9]
+        attribute :delivered_quantity, Types::Sized::Float[-9_999_999_999_999.9, 9_999_999_999_999.9]
 
         # Description Description Row description. 50 characters
         attribute :description, Types::Sized::String[50]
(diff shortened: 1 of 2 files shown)


ehannes:
Hi @mbirman!
This is (soon was) a know problem and on our TODO list, so many thanks for your contribution! We'll take a look at it.

Great to hear you are using our gem. It is currently under heavy development. Would you like to join our slack?

ehannes:
@d-Pixie Rubocop is complaining: `Example has too many lines [6/5]`, but in this case it's not much to do  (mostly because of the long `9_999_999_999_999.9`). Maybe we should just accept this and add an exception in CC? (I don't have the privileges, so it's up to you :smile_cat:)

d-Pixie:
Welcome and thanks for the contribution. Fortnox documentation on exactly what is allowed is sometimes missing or unclear. We are happy to correct it though 😉

d-Pixie:
@mbirman I will release a new version of the gem with this change tomorrow. Look for version 0.6.2 😊

mbirman:
Wow guys, you're fast 👍 

@ehannes yes, I'd love to! Do you need my email or something?

ehannes:
Concerns #95

PR #151, opened 2018-03-09
Add specs for CircularQueue (#134)

mbirman:


d-Pixie:
You are testing this in a very implementation dependant way. If we change how the queue is implemented all the tests will break :)

The only public method is `#next` so that is what we should test. What can we assert about the queue based on only that entry point?

We know that it should work with the base case of only one element, so given `q = Fortnox::API::CircularQueue.new(1)` any call to `q.next` should yield `1`.

We know it should loop over the elements of a given array in order, so given `q = Fortnox::API::CircularQueue.new(1,2,3)` subsequent calls to `q.next` should yield different results three times in a row, without repeat.

We also need to test that it actually repeats in order over multiple values. Give the random start index this will require us to find out what index it started out before we can test, but setting up a queue with unique items, getting the first value and its index in the input and then ensuring that it walks a complete loop in order over the items isn't that complicated.

The random start index is a bit hard to test, since it is random and any test could potentially be flaky due to that, but we can try. Given `q1 = Fortnox::API::CircularQueue.new((1..1_000_000).to_a)` and `q2 = Fortnox::API::CircularQueue.new((1..1_000_000).to_a)` the chance of `q1.next == q2.next` is literally one in a million, and I think we can live with having a test suit that flakes one time in a million.

These tests should test all properties of the queue that we care about using only the public API, which is what we want :)

ehannes:
I think we want something like this @mbirman. It can probably be more thought through, but I think you'll get the point. Do you want to give it a try?
```
describe CircularQueue do
  describe 'start index' do
    context 'when running several times' do
      it 'does not start with the same item each time'
    end
  end

  describe '#next' do
    # you could probably mock the random_start_index to make tests easier here

    context 'when called the first time' do
      it 'returns the first item'
    end

    context 'when called twice' do
      it 'returns the second item'
    end

    context 'when called list size numer of times' do
      it 'returns the first item'
    end 
  end
end
```

ehannes:
Replaced by #164

d-Pixie:
@mbirman If you want to see how we did it have a look at #164 🙂

PR #158, opened 2018-06-04
Cast AccountNumber string to integer

samuel02:
According to the documentation at https://developer.fortnox.se/documentation/resources/customers/
the field `SalesAccount` on the `Customer` is a 4 digit string.

### DIFF
--- a/lib/fortnox/api/types.rb
+++ b/lib/fortnox/api/types.rb
@@ -38,6 +38,7 @@ module Fortnox
       AccountNumber = Strict::Int
                       .constrained(gteq: 0, lteq: 9999)
                       .optional
+                      .constructor(Dry::Types::Coercions::Form.method(:to_int))
 
       ArticleType = Strict::String
                     .constrained(included_in: ArticleTypes.values)
--- a/lib/fortnox/api/types.rb
+++ b/lib/fortnox/api/types.rb
@@ -38,7 +38,15 @@ module Fortnox
       AccountNumber = Strict::Int
                       .constrained(gteq: 0, lteq: 9999)
                       .optional
-                      .constructor(Dry::Types::Coercions::Form.method(:to_int))
+                      .constructor do |input|
+                        unless input.nil? || input.to_s.empty?
+                          Integer(input)
+                        else
+                          input
+                        end
+                      rescue ArgumentError
+                        input
+                      end
 
       ArticleType = Strict::String
                     .constrained(included_in: ArticleTypes.values)
--- a/spec/fortnox/api/types/account_number_spec.rb
+++ b/spec/fortnox/api/types/account_number_spec.rb
@@ -14,16 +14,26 @@ describe Fortnox::API::Types do
   end
 
   context 'when AccountNumber created with valid number' do
-    subject { klass['1234'] }
-
-    it { is_expected.to eq 1234 }
+    include_examples 'equals input', 1234
   end
 
   context 'when AccountNumber created with a too large number' do
-    include_examples 'raises ConstraintError', '10000'
+    include_examples 'raises ConstraintError', 10_000
   end
 
   context 'when AccountNumber created with a negative number' do
-    include_examples 'raises ConstraintError', '-1'
+    include_examples 'raises ConstraintError', -1
+  end
+
+  context 'when AccountNumber created with an invalid string' do
+    include_examples 'raises ConstraintError', 'foo'
+  end
+
+  context 'when AccountNumber created with valid string' do
+    subject { klass['1234'] }
+
+    it 'casts it to a number' do
+      is_expected.to eq 1234
+    end
   end
 end
--- a/spec/fortnox/api/types/account_number_spec.rb
+++ b/spec/fortnox/api/types/account_number_spec.rb
@@ -13,19 +13,17 @@ describe Fortnox::API::Types do
     it { is_expected.to be_nil }
   end
 
-  context 'when AccountNumber created with empty string' do
-    include_examples 'raises ConstraintError', ''
-  end
-
   context 'when AccountNumber created with valid number' do
-    include_examples 'equals input', 1234
+    subject { klass['1234'] }
+
+    it { is_expected.to eq 1234 }
   end
 
   context 'when AccountNumber created with a too large number' do
-    include_examples 'raises ConstraintError', 10_000
+    include_examples 'raises ConstraintError', '10000'
   end
 
   context 'when AccountNumber created with a negative number' do
-    include_examples 'raises ConstraintError', -1
+    include_examples 'raises ConstraintError', '-1'
   end
 end
--- a/spec/fortnox/api/types/account_number_spec.rb
+++ b/spec/fortnox/api/types/account_number_spec.rb
@@ -13,6 +13,10 @@ describe Fortnox::API::Types do
     it { is_expected.to be_nil }
   end
 
+  context 'when AccountNumber created with empty string' do
+    include_examples 'raises ConstraintError', ''
+  end
+
   context 'when AccountNumber created with valid number' do
     include_examples 'equals input', 1234
   end


lib/fortnox/api/models/customer.rb
```diff
@@ -161,7 +161,7 @@ class Customer < Fortnox::API::Model::Base
         attribute :project, Types::Nullable::String
 
         # SalesAccount	Sales account of the customer, 4 digits
-        attribute :sales_account, Types::AccountNumber
+        attribute :sales_account, Types::Sized::String[4]
```
d-Pixie: This is not ok as is as it would allow the account NUMBER to be set to `"abcd"` etc. If the API actually barfs if we send it a number here we should do some casting instead.
Is this something you have experienced problems with or was it an attempt to prevent problems? Also note that we do NOT reflect the Fortnox API spec directly in our API due to theirs being full of errors, conflicts and inconsistencies. Instead we try to normalise the API as much as possible, returning actual Ruby Date objects for date and time, always representing numeric types as numbers, using enums in place of strings for countries etc.

d-Pixie (changes requested):

samuel02:
@d-Pixie @ehannes I updated this PR to instead cast it from a string.

spec/fortnox/api/types/account_number_spec.rb
```diff
@@ -13,19 +13,17 @@
     it { is_expected.to be_nil }
   end
 
-  context 'when AccountNumber created with empty string' do
```
ehannes: @samuel02 Why dropping this case? Is it not needed anymore?

spec/fortnox/api/types/account_number_spec.rb
```diff
@@ -13,19 +13,17 @@
...
-    include_examples 'raises ConstraintError', ''
-  end
-
   context 'when AccountNumber created with valid number' do
-    include_examples 'equals input', 1234
+    subject { klass['1234'] }
```
ehannes: `1234` should still be a valid number, right? We want to test both string and integer input. So we should probably add a `when AccountNumber created with valid string` case here instead of changing this integer to string.

spec/fortnox/api/types/account_number_spec.rb
```diff
@@ -13,19 +13,17 @@
...
+
+    it { is_expected.to eq 1234 }
   end
 
   context 'when AccountNumber created with a too large number' do
-    include_examples 'raises ConstraintError', 10_000
```
ehannes: Same here. We should probably keep these as integers. We only need to test the casting once.

spec/fortnox/api/types/account_number_spec.rb
```diff
@@ -13,19 +13,17 @@
...
+    include_examples 'raises ConstraintError', '10000'
   end
 
   context 'when AccountNumber created with a negative number' do
-    include_examples 'raises ConstraintError', -1
```
ehannes: Same here. We should probably keep these as integers. We only need to test the casting once.

ehannes (changes requested):
Nice! Please have a look my comments. Feel ask if something is unclear or if I have misunderstood something.

spec/fortnox/api/types/account_number_spec.rb
samuel02: It got complicated :) The problem is that the constructor `Dry::Types::Coercions::Form.to_int` will return `nil` in the case of an empty string. And since the type is declared as optional that's ok and it won't raise anything. What do you think? Should I write a custom constructor that handles that case as well?

d-Pixie: Well, it being complicated is hardly a reason to not test for it. In fact it is a very good reason TO test for it! 😄

lib/fortnox/api/types.rb
```diff
@@ -38,6 +38,7 @@ module Types
       AccountNumber = Strict::Int
                       .constrained(gteq: 0, lteq: 9999)
                       .optional
+                      .constructor(Dry::Types::Coercions::Form.method(:to_int))
```
d-Pixie: I like the thought behind this. You can probably achieve the same result by changing `Strict::Int` to `Coercible::Int` at the top of the declaration. I'm assuming this doesn't work for the empty string case though.

To correctly deal with nil and empty string you can do something like this:

```
AccountNumber = Strict::Int
  .constrained(gteq: 0, lteq: 9999)
  .optional
  .constructor { |v| Integer(v) unless v.nil? or v.empty? }
```

d-Pixie (changes requested):

spec/fortnox/api/types/account_number_spec.rb
samuel02: Sure :) My thinking was just that it doesn't maybe need to be *that* strict. The consequence of the new code would be that an empty string would end up as a nil value which was already permitted. So in practice it probably doesn't matter that much. But sure :) I pushed another implementation. Unfortunately it's not enough with the constructor you provided since `Integer` doesn't implement `#empty?`, if we'd have `ActiveSupport` we could have used `Object#blank?` but I didn't want to pull that in. So I ended up writing my own. WDYT?

lib/fortnox/api/types.rb
```diff
@@ -38,6 +38,15 @@ module Types
       AccountNumber = Strict::Int
                       .constrained(gteq: 0, lteq: 9999)
                       .optional
+                      .constructor do |input|
+                        unless input.nil? || input.to_s.empty?
```
ehannes: Are these cases tested?

lib/fortnox/api/types.rb
```diff
@@ -38,6 +38,15 @@ module Types
       AccountNumber = Strict::Int
                       .constrained(gteq: 0, lteq: 9999)
                       .optional
+                      .constructor do |input|
```
ehannes: This constructor is now 9 lines. Can we create a separate method for this? I'm pretty sure Rubocop will complain about this :)

lib/fortnox/api/types.rb
```diff
@@ -38,6 +38,15 @@ module Types
...
+                        unless input.nil? || input.to_s.empty?
+                          Integer(input)
+                        else
+                          input
+                        end
+                      rescue ArgumentError
+                        input
```
ehannes: Is this case tested?

ehannes (changes requested):
Now we're close :)

spec/fortnox/api/types/account_number_spec.rb
```diff
@@ -28,4 +28,16 @@
...
   end
+
+  context 'when AccountNumber created with an invalid string' do
+    include_examples 'raises ConstraintError', 'foo'
+  end
+
+  context 'when AccountNumber created with valid string' do
+    subject { klass['1234'] }
+
+    it 'casts it to a number' do
```
d-Pixie: We don't really care HOW it does something here, just that we get the correct result. So this could be:
```ruby
context 'created with valid string' do
  subject { klass['1234'] }
  it { is_expected.to eq 1234 }
end
```
Utilizing the readability of RSpecs one line syntax.

d-Pixie: Actually if we consider how we do things we can change it to `'created with valid value that responds to to_s'`, since that is actually what we allow.

spec/fortnox/api/types/account_number_spec.rb
```diff
@@ -28,4 +28,16 @@
   context 'when AccountNumber created with a negative number' do
     include_examples 'raises ConstraintError', -1
   end
+
+  context 'when AccountNumber created with an invalid string' do
```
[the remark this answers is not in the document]
ehannes: Yes, it should be `describe Fortnox::API::Types::AccountNumber`. My bad!

lib/fortnox/api/types.rb
d-Pixie: Not all valid/invalid cases are tested. We could test for everything but right now, with my suggested changes, we do test the two main cases: Things that respond to to_s and has a valid value will convert, the rest will cause a Dry types constraint error...

ehannes:
Closing this as it's stale and obsolete now.

PR #166, opened 2018-10-30
Fixes #103

### LINKED ISSUE #103: Httparty version is currently locked down
Httparty version `0.15.0` breaks gem. The version is temporarily locked down to `0.14.0` until we find a fix. See issue 547 in httparty's repo for more information. See branch `ehannes:httparty-issue` for test branch.

ehannes:
@d-Pixie I'm not sure this is a correct fix for this issue. See https://github.com/jnunemaker/httparty/issues/547


d-Pixie:
Hmmm, this looks broken. Might it be because I had the worn remote? But it should warn me then, shouldn't it? ...
I'll fix it.

d-Pixie:
Better ...

There are two failing tests now. They are due to the fact that I moved where we get the client secret from initialisation to call. So the tests won't trigger the failure condition as written, you need to make a call to somewhere to get the error.

ehannes:
@d-Pixie I have fixes the broken test now.

But we should probably throw away all our VCR cassettes and record them again before we know this actually works.

PR #169, opened 2019-04-10
Spec fix

ehannes:
Fixes broken specs. I just ran `bundle install` and a lot of tests were broken. This has to do with updates to `Dry`. Though, note that we have not updates our dependency locks on the dry gems.

### DIFF
--- a/spec/fortnox/api/repositories/base_spec.rb
+++ b/spec/fortnox/api/repositories/base_spec.rb
@@ -5,19 +5,33 @@ require 'fortnox/api'
 
 describe Fortnox::API::Repository::Base do
   using_test_class do
+
     module Model
-      class Test
+      class RepositoryBaseTest
       end
     end
     module Repository
       class Test < Fortnox::API::Repository::Base
-        MODEL = Model::Test
+        MODEL = Model::RepositoryBaseTest
       end
+    end
+  end
+
+  before do
+    begin
+      # NOTE: This registry key is leaking throughout the whole test suite...
+      # I don't know of any way to remove a key from the registry without resetting the whole registry.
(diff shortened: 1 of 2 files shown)


spec/fortnox/api/repositories/base_spec.rb
```diff
@@ -5,19 +5,33 @@
...
+        MODEL = Model::RepositoryBaseTest
       end
     end
+  end
+
+  before do
+    begin
+      # NOTE: This registry key is leaking throughout the whole test suite...
+      # I don't know of any way to remove a key from the registry without resetting the whole registry.
+      Fortnox::API::Registry.register(:repositorybasetest, Model::RepositoryBaseTest)
```
d-Pixie: The hacky way of doing this is to reach into the concurrent hash and delete the key:
```
Fortnox::API::Registry._container.delete( :repositorybasetest ) 
```

ehannes: This code:
```
  before do
    Fortnox::API::Registry._container.delete( :repositorybasetest )
    Fortnox::API::Registry.register(:repositorybasetest, Model::RepositoryBaseTest)
  end

```

Gives this error:
```
      Dry::Container::Error:
        There is already an item registered with the key "repositorybasetest"
```

d-Pixie: What do you get back from the `Fortnox::API::Registry._container.delete( :repositorybasetest )` call? It should return the stored value. And after, can you still get the stored value?

Can we solve this by changing the name of the key? Can you dynamically use the test name in RSpec for the key name? Sure, it will leave a lot of crap in the registry during the test run, but it is a pragmatic solution.

ehannes: It was stored as a string, not a symbol :)

Issue #153, opened 2018-03-27
Invalid validation for Invoice and Order Country


ehannes:
When we wrote the validations for `Invoice` and `Order`, we assumed Fortnox API uses country codes for the `Country` attribute. Unfortunately, that is not true. `Country` is a string containing the full country name! If you look at the documentation examples, you see that they have `"Country": "Sverige"`, and I recently got this error in one of our apps when a user tried to fetch an Order from Fortnox API: `[Fortnox::API::Model::Order.new] "United Kingdom" (String) has invalid type for :country violates constraints (included_in?(["AF", "AX", "AL", "DZ", "AS", "AD", "AO", "AI"...`

If you read the documentation for `Invoice`, it says that the string _"Must be a name of an existing country."_. What that means I do not know, so I've send a question to Fortnox API support. Apparently, you can write the country name in both Swedish (as the documentation exampel) and English (as the example above). What other languages are supported? Why do they not use country codes? There are so many questions... :confused:  :question:

ehannes:
`Sverige` is ok:

```
pry(main)> Fortnox::API::Repository::Order.new.save(Fortnox::API::Model::Order.new(customer_number: '1', country: 'Sverige', order_rows: [Fortnox::API::Types::OrderRow.new(ordered_quantity: 1, article_number: '0000')]))
[...]

PR #170, opened 2019-05-29
Fix 153 country validation

ehannes:
Adds proper validation for Country attribute. Read more in #153.

Fortnox are not supporting country codes, you must use the name of the country. And it's not the official name. I found the correct version of the country name in the `translations` map:
```
ISO3166::Country['GB'].translations['en'] 
=> "United Kingdom"
```

Sweden is a special case. Fortnox API does NOT support the English version of the country name.

All countries except `Sverige` is in English. We could support Swedish country names if we want (just by calling `translations['se']` instead, but I don't see the point in that since the customer is obviously foreign. English seems like a better choice.

Right now we are supporting country codes, the country name in English and in Swedish. You could support more languages and move that into a configuration if we would like to.

I have added integration tests in the `Invoice` repository test.


lib/fortnox/api/types.rb
```diff
@@ -45,9 +48,24 @@ module Types
...
+                    .constructor do |value|
+                      next value if value.nil? || value == ''
+
+                      # Fortnox only supports Swedish translation of Sweden
+                      sweden_variants = ['se', :se, 'sweden', :sweden, 'sverige', :sverige]
+
+                      next CountryCodeString.new('SE') if sweden_variants.include?(value.downcase)
+
+                      country = ::ISO3166::Country[value] ||
+                                ::ISO3166::Country.find_country_by_name(value.to_s.downcase) ||
```
d-Pixie: This only works on English names and `find_country_by_translated_names` works on names given in any of the locales that you have configured, if I understand it correctly. But this data is from Fortnox that only permits names in English, unless it is Sverige which we already handle above. So the last clause here should never get called?

Also `find_country_by_name` works fine with both strings and symbols regardless of case, just like you already call `find_country_by_translated_names` 🙂

lib/fortnox/api/types/fortnox_country_code.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
```
d-Pixie: This file has the wrong name. Also maybe put this in `lib/fortnox/api/types/shims/country_code_string.rb` just to underline that it is a special type of type? 🙂

spec/fortnox/api/mappers/base_spec.rb
```diff
@@ -111,57 +117,19 @@
...
-    end
-  end
-
-  describe 'VATType' do
-    include_examples 'identity mapper', :vat_type do
-      let(:value) { Fortnox::API::Types::VATType['SEVAT'] }
+        it 'translates code to country name in Swedish' do
+          is_expected.to eq('Sverige')
+        end
+      end
```
d-Pixie: Missing tests for empty string, `nil` and random crap 🙂

spec/fortnox/api/types/country_code_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        it "converts \"#{sweden_input}\" to \"SE\"" do
+          expect(described_class[sweden_input]).to eq 'SE'
+        end
+      end
+    end
+  end
+
+  context 'with invalid input' do
+    invalid_inputs = [
+      'SEA', 'S', 'nonsense', :s
```
d-Pixie: Ah, this is where you should test with `nil` and empty string then? They are special cases in the implementation so ...

spec/fortnox/api/mappers/base_spec.rb
d-Pixie: Ok, found them further down, but that makes me wonder what we are testing here? If you want to see if we have a mapper for the type registered you should just test that `Fortnox::API::Registry[:countrycodestring]` is callable, you should not really call it.

In this case the function is unlikely to change over time, so this way of testing would hold in this special case, but in another case this test would break if the _functionality_ of the registered object changes, even if the mapper is still correctly registered. So you are testing too much in this test.

d-Pixie:
Despite the above comments: Fantastic job on that refactor! Code is a lot better after. I like everything about how this ended up 😄

lib/fortnox/api/types.rb
```diff
@@ -45,9 +48,24 @@ module Types
...
-                    .constrained(included_in: CountryCodes.values)
                     .optional
-                    .constructor(EnumConstructors.sized(2))
+                    .constructor do |value|
+                      next value if value.nil? || value == ''
+
+                      # Fortnox only supports Swedish translation of Sweden
+                      sweden_variants = ['se', :se, 'sweden', :sweden, 'sverige', :sverige]
+
+                      next CountryCodeString.new('SE') if sweden_variants.include?(value.downcase)
```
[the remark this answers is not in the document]
ehannes: This also works with both strings and symbols: https://ruby-doc.org/core-2.2.0/Symbol.html#method-i-downcase

lib/fortnox/api/types.rb
ehannes: No. Fortnox support both English and Swedish country names, except for one case: the country name of Sweden must be in Swedish.

Good catch! I'll remove the `.to_s.downcase` :)

lib/fortnox/api/types/fortnox_country_code.rb
ehannes: Correct, good catch! Should we keep this class or should we override the `class` method as we talked about?

spec/fortnox/api/mappers/base_spec.rb
ehannes: We need this test to see that our mapper is registered with the correct name. It was when I wrote this test that I found out that a lot of our past mappers were unused because they were registered with name that were never looked up. But you are correct, we should not call the mapper here.

ehannes:
I like your comments, it makes the code better and I continue to learn  :)

spec/fortnox/api/mappers/base_spec.rb
ehannes: And you just found two bugs @d-Pixie! Great review!

spec/fortnox/api/types/country_code_spec.rb
ehannes: Great that you pointed this out! But both `nil` and empty string is valid input, so they should be tested above in the code :)

lib/fortnox/api/types/fortnox_country_code.rb
d-Pixie: I think we can keep the class, it is cleaner - even if the other one is more fun 🙂

PR #175, opened 2019-07-03
New version

ehannes:
It's time to release a new version. Change included:

# Fixes
Country attribute for Invoice and Order is now validated correctly.


ehannes:
Will open a new PR to see if I can get Code Climate statuses to work

PR #176, opened 2019-07-04
Version 0.7.0

ehannes:
It's time to release a new version. From our new changelog:

# Added
- Adds build test for Ruby 2.6.0, 2.6.3 and 2.7.0-preview.

# Removed
- Drops support for Ruby < 2.4.0 since they are deprecated

# Fixed
- Country attribute for Invoice and Order is now validated correctly.
- Fixes deprecation warnings


ehannes:
@d-Pixie we should add all releases to GitHub. There are several missing ones. Compare the releases at https://rubygems.org/gems/fortnox-api with https://github.com/accodeing/fortnox-api/releases

PR #186, opened 2019-09-30
Fixes incorrect validation for AccountNumber

### LINKED ISSUE #183: AccountNumber validation
Our validation of `AccountNumber` is invalid. It is used by the `Customer` model and maybe some more resource that I do not know of.

Fortnox allows two `unset` values for `AccountNumber`: `''` and `null`. We only support the latter for the moment. So when we get a `Customer` that has the attribute set to empty string, we get this error message:
```
"" (String) has invalid type for :sales_account violates constraints (type?(Integer, "") AND gteq?(0, "") AND lteq?(9999, "") failed) (Dry::Types::SchemaError)
```

So we need to change our validation logic to support this case.

ehannes:
Fixes #183 
This was merged to `development` some time ago, but it has not yet reached `master` :)


d-Pixie:
Sjukt bra att du alltid uppdaterar testerna Hannes 😄

ehannes:
If I didn't update the tests, I would never dare to merge anything. I do too many mistakes, so I need the tests, just like all developers do ;)

Issue #190, opened 2020-12-21
Add support for new ROT/RUT functionality


ehannes:
Fortnox updated their endpoints with breaking changes last week. More information is available here: [Updates in Tax Reduction](https://developer.fortnox.se/blog/updates-in-tax-reduction/Updates). The gem does currently not support this new functionality, so creating orders and invoices with housework types is not working correctly now, at least I think that's the case.

Relevant info for us:
- A new property called `TaxReductionType` will be added to the Invoice endpoint. It is a String that must be `rot`, `rut`, `green`, `none` or `null`  that decides which housework type the invoice will contain. **`TaxReductionType` is Required if the invoice row has `HouseWork: True`**
- An invoice with housework must be in the currency SEK.
- Old invoices will have `taxReductionType` set to `null`
- New validations on which types of housework that will be able to combine in the external API.
- The response acts the same way for the endpoints contract, invoice, offer and order.
- You can not mix types of housework between rot, rut or green.
- You can not have a row with `HouseWorkType` if you haven’t chosen a `taxReductionType`.
- New types of `HouseWorkType`
- It should not be possible to ROT and RUT housework types, and they should not be able to have a RUT housework type if you set the tax reduction type to ROT.
- If the field `HouseWork` in the Invoice resource is true then the field `HouseWorkType` under every `InvoiceRow` has to be set even if `HouseWork` is false on the invoice row.

There's also a new `HouseWorkType` called `EMPTYHOUSEWORK`
- It is not possible to use `EMPTYHOUSEWORK` in combination with the following houseworktypes `TRANSPORTATIONSERVICES` and `WASHINGANDCAREOFCLOTHING`. I don't think we can implement a check for that right now though.
- It can be used with invoice rows with both negative and positive amount.

ehannes:
We won't follow address the following:
- An invoice with housework must be in the currency SEK.
- You can not mix types of housework between rot, rut or green.
- You can not have a row with `HouseWorkType` if you haven’t chosen a `taxReductionType`.
- It should not be possible to ROT and RUT housework types, and they should not be able to have a RUT housework type if you set the tax reduction type to ROT.
- If the field HouseWork in the Invoice resource is true then the field `HouseWorkType` under every `InvoiceRow` has to be set even if `HouseWork` is false on the invoice row.
[...]

PR #194, opened 2020-12-22
Fix 190 adds support for tax reduction type

ehannes:
- Adds support for `TaxReductionType`
- Restructures house work types enums
- Adds new house work types
- Adds TODO for new house work types 2021-01-01


spec/fortnox/api/types/housework_types_spec.rb
```diff
@@ -46,23 +49,135 @@
...
+            article_number: '0000',
+            housework_type: Fortnox::API::Types::HouseworkTypes['CONSTRUCTION'],
+            housework: true
+          )
+        ]
+      )
+    end
+
+    let(:error_message) do
+      "Dokument med skattereduktionstypen 'rut' får inte innehålla rader med husarbetestypen 'CONSTRUCTION'."
```
d-Pixie, line 175: You are using a hardcoded part of the message here. If you change the test to use a different `Fortnox::API::Types::HouseworkTypes` the message will change and the test will fail. So you should probably interpolate the expected value, `CONSTRUCTION`, here.

d-Pixie (approved):

ehannes:
Fixes #190

PR #197, opened 2021-09-01
Deprecates Ruby 2.5 and updates Travis Ruby targets

ehannes:
Time to update some Ruby versions. 2.5 has reached end of life and we should start testing against Ruby 3 :)


ehannes:
Something is wrong with Travis. I'll close this and try to open a new.

PR #199, opened 2021-09-01
New version

ehannes:
Time to update some Ruby versions. 2.5 has reached end of life and we should start testing against Ruby 3 :)


ehannes:
Yay, at last, we have some feedback from Travis. Unfortunately only from one of the two checks...

Apparently, the plugin isn't ready for Ruby 3 yet since that build fails. I vote for merging this PR and add a separate ticket for adding support for Ruby 3.

d-Pixie (approved):

ehannes:
The build fails on Ruby 3.0:
```
$ rvm get head
Downloading https://get.rvm.io
Could not download rvm-installer, please report to https://github.com/rvm/rvm/issues
The command "rvm get head" failed and exited with 60 during .
```

d-Pixie (approved):

PR #207, opened 2022-04-27
Fixes #206 increases row description limit

### LINKED ISSUE #206: Update validation for document row description limit
Previously, the limit for `description` in a document row was 50. Now it has been changed to 255.

ehannes:
Fixes #206

### DIFF
--- a/lib/fortnox/api/types/document_row.rb
+++ b/lib/fortnox/api/types/document_row.rb
@@ -22,8 +22,8 @@ module Fortnox
         # DeliveredQuantity Delivered quantity. 14 digits
         attribute :delivered_quantity, Types::Sized::Float[-9_999_999_999_999.9, 9_999_999_999_999.9]
 
-        # Description Description Row description. 50 characters
-        attribute :description, Types::Sized::String[50]
+        # Description Description Row description.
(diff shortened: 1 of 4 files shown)


lib/fortnox/api/types/document_row.rb
```diff
@@ -22,8 +22,8 @@ class DocumentRow < Fortnox::API::Types::Model
         # DeliveredQuantity Delivered quantity. 14 digits
         attribute :delivered_quantity, Types::Sized::Float[-9_999_999_999_999.9, 9_999_999_999_999.9]
 
-        # Description Description Row description. 50 characters
-        attribute :description, Types::Sized::String[50]
+        # Description Description Row description.
```
d-Pixie, line 25: I guess it is ok to remove it from the comment - but then we should really remove it from all comments just to be consistent 🙂

d-Pixie (approved):

lib/fortnox/api/types/document_row.rb
ehannes: Ye, precicely. We should remove all comments. I can do that in a separate commit :)

PR #216, opened 2022-05-03
Fix 214 sales account

### LINKED ISSUE #214: Fix Customer SalesAccount type
Apparently, `SalesAccount` attribute on `Customer` is a string, not an integer as the `Number` attribute in `Account` is.

According to the Fortnox documentation, in `Customer` a `Sales Account` is a "string 4 characters", but in `Account` the `Numer` attribute is an "integer <int32> [ 1000 .. 9999 ]". So Fortnox has different data types for "the same thing" on different endpoints :weary: 

We are currently using the same implementation for validating both these attributes, but that's not a correct implementation. We get this error:

```
[Fortnox::API::Model::Customer.new] "3010" (String) has invalid type for :sales_account violates constraints
(type?(Integer, "3010") AND gteq?(0, "3010") AND lteq?(9999, "3010") failed) (Fortnox::API::AttributeError)
```

This is how the attribute is defined now, which is obviously wrong:
```
# SalesAccount  Sales account of the customer, 4 digits
attribute :sales_account, Types::AccountNumber
```

ehannes:
Fixes #214. See the issue for more information. The Fortnox API is just stupid here...

### DIFF
--- a/lib/fortnox/api/models/customer.rb
+++ b/lib/fortnox/api/models/customer.rb
@@ -160,7 +160,7 @@ module Fortnox
         # Project  Project of the customer, Project in Fortnox
         attribute :project, Types::Nullable::String
 
-        # SalesAccount  Sales account of the customer, 4 digits
+        # SalesAccount  Sales account of the customer, 4 characters
         attribute :sales_account, Types::SalesAccount
 
         # ShowPriceVATIncluded  Show prices with VAT included or not
--- a/lib/fortnox/api/models/customer.rb
+++ b/lib/fortnox/api/models/customer.rb
@@ -161,7 +161,7 @@ module Fortnox
         attribute :project, Types::Nullable::String
 
         # SalesAccount  Sales account of the customer, 4 digits
-        attribute :sales_account, Types::AccountNumber
+        attribute :sales_account, Types::SalesAccount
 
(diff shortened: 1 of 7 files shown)


lib/fortnox/api/models/customer.rb
```diff
@@ -161,7 +161,7 @@ class Customer < Fortnox::API::Model::Base
         attribute :project, Types::Nullable::String
 
         # SalesAccount  Sales account of the customer, 4 digits
-        attribute :sales_account, Types::AccountNumber
+        attribute :sales_account, Types::SalesAccount
```
d-Pixie, line 164: Comment needs updating

d-Pixie (approved):

lib/fortnox/api/models/customer.rb
ehannes: We should probably just get rid of these comments and refer to the documentation instead... These comments will just get old and I don't see the point in keeping them updated. Do you @d-Pixie?

d-Pixie: No, I agree. I almost wrote " ... or removed" yesterday 😄

ehannes: I'll change the comment for this PR so we leave this in a correct state.

PR #220, opened 2022-05-23
WIP: Feature/205 auth2 support

ehannes:


README.md
```diff
@@ -104,47 +104,91 @@ $ gem install fortnox-api
 
 # Usage
 
-## Getting an AccessToken
-To make calls to the API server you need a `ClientSecret` and an `AccessToken`. When you sign up for an API-account with Fortnox you should get a client secret and an authorization code. To get the access token, that is reusable, you need to do a one time exchange with the API-server and exchange your authorization code for an access token. For more information about how to get access tokens, see Fortnox developer documentation.
+## Authorization
+> :warning: Before 2022, Fortnox used a client ID and a fixed access token for authorization. This way of is now deprecated. The old access tokens have a life span of 10 years according to Fortnox. They can still be used, but you can't issue any new long lived tokens and they recommend to migrate to the new authorization process. This gem will no longer support the old way of authorization.
```
d-Pixie, line 108: Add note on last version to support the old access tokens.

README.md
```diff
@@ -104,47 +104,91 @@ $ gem install fortnox-api
...
 
-## Configuration
-To configure the gem you can use the `configure` block. A `client_secret` and `access_token` (or `access_tokens` in plural, see [Multiple AccessTokens](#multiple-accesstokens)) are required configurations for the gem to work so at the very minimum you will need something like:
+You need to have a Fortnox app and to create such an app, you need to register as a Fortnox developer. It might feel as if "I just want to create an integration to Fortnox, not build a public app to in the marketplace". Ye, we agree... You don't need to release the app to the Fortnox Marketplace, but you to have that Fortnox app. Also, see further Fortnox app requirements down below.
```
d-Pixie, line 110: Ye -> Yeah

d-Pixie: "You don't need to release the app to the Fortnox Marketplace" -> "You don't need to release the app *on* the Fortnox Marketplace"

d-Pixie: "but you to have that Fortnox app" ???

README.md
```diff
@@ -104,47 +104,91 @@ $ gem install fortnox-api
...
 
-### Multiple AccessTokens
+And could then be used like this:
 
-Fortnox uses quite low [API rate limits](https://developer.fortnox.se/blog/important-implementation-of-rate-limits/). The limit is for each access token, and according to Fortnox you can use as many tokens as you like to get around this problem. This gem supports handeling multiple access tokens natively. Just set the `access_tokens` (in plural, compared to `access_token` that only takes a String) to a list of strings:
 ```ruby
-  config.client_secret = 'P5K5wE3Kun'
-  config.access_tokens ['a78d35hc-j5b1-ga1b-a1h6-h72n74fj5327', 's2b45f67-dh5d-3g5s-2dj5-dku6gn26sh62']
+  config.storage = MyStorage.new(
```
d-Pixie, line 148: Wouldn't this be:
```
  config.storage = MyStorage.new
end
```
I would expect the implementation of the storage to handle any loading from env on init etc ...

README.md
```diff
@@ -104,47 +104,91 @@ $ gem install fortnox-api
...
+
+### Support for multiple Fortnox accounts
+Yes, we support working with several accounts at once. Simply expose `token_stores` from the storage and set it to a hash where the keys (called *token store*) represents different Fortnox accounts and provide a refresh token for each such account. If you provide a `:default` token store, this is used as default by all repositories.
 
 ```ruby
+class MyStorage
+  attr_accessor :token_stores
+
+  def initialize(token_stores)
+    @token_stores = token_stores
```
d-Pixie, line 165: You don't need `@token_stores` since you use `attr_accessor`, just `token_stores` is fine (and recommended since it protects you from misspelling the variable name).
So this implementation is completely divorced from the one for a single token? I would expect this to expose a set of `MyStorage` compatible stores. Just for API consistency. Or change the other implementation to act as a hash instead.

README.md
```diff
@@ -104,47 +104,91 @@ $ gem install fortnox-api
...
 
 Fortnox::API::Repository::Customer.new # Using token store :default
 Fortnox::API::Repository::Customer.new( token_store: :another_account ) # Using token store :another_account
 ```
-The tokens per store are rotated between calls to the backend as well. That way you can create a web app that connects to multiple Fortnox accounts and uses multiple tokens for each account as well.
+
+Note that when you provide multiple token stores, you need to keep *all those refresh tokens* alive.
+
+### Multiple access tokens
+As of november 2021 and the new OAuth 2 flow, Fortnox has made [adjustments to the rate limit](https://developer.fortnox.se/blog/adjustments-to-the-rate-limit/) and it is no longer calculated per access token (if you are not using the old auth flow, but that flow is no longer supported by this gem).
```
d-Pixie, line 191: "but that flow is no longer supported by this gem as of vX.X.X"

bin/get_tokens
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "base64"
+require 'dotenv'
+require 'httparty'
+
+Dotenv.load('.env.test')
```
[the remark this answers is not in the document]
d-Pixie: Doesn't dotenv load files in a specific order, so that `.ev.test` would override any `.env` that was there ...
Also the `.env.test` file should be committed to the repo. I seem to remember from above that it was added to `.gitignore`.

bin/get_tokens
```diff
@@ -0,0 +1,75 @@
...
+
+print "Let's get you an access token and a refresh token for Fortnox API."
+
+client_id = ENV.fetch('FORTNOX_API_CLIENT_ID')
+client_secret = ENV.fetch('FORTNOX_API_CLIENT_SECRET')
+redirect_uri = ENV.fetch('FORTNOX_API_REDIRECT_URI')
+scopes = ENV.fetch('FORTNOX_API_SCOPES')
+credentials = "#{Base64.encode64("#{client_id}:#{client_secret}")}"
+oauth_endpoint = 'https://apps.fortnox.se/oauth-v1'
```
d-Pixie, line 17: Constant?

bin/get_tokens
```diff
@@ -0,0 +1,75 @@
...
+
+def exit_if_empty(variable, message)
+  if variable.empty?
+    print "\n#{message}"
+    exit
+  end
+end
+
+exit_if_empty(client_id, 'Missing client ID! Exiting.')
```
d-Pixie, line 26: `raise "Missing client ID! Exiting." if client_id.empty?`

lib/fortnox/api/repositories/base.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require 'httparty'
+require 'jwt'
+require "base64"
```
d-Pixie, line 5: single quote

lib/fortnox/api/repositories/base.rb
```diff
@@ -88,20 +80,105 @@ def base_url
...
           Fortnox::API.config
         end
 
-        def token_store_not_found!(store_name)
-          raise MissingConfiguration,
-                "There is no token store named #{store_name}. Available stores are #{config.token_store.keys}."
+        def get_token_store(name)
+          storages = config.storages
+
+          return validate_store(storages[name], name) unless storages.nil?
```
d-Pixie, line 102:
```
storage = storages[name] || config.storage
validate_store(storage, name)
```

lib/fortnox/api/request_handling.rb
```diff
@@ -16,6 +16,13 @@ def raise_api_error(error, response)
       def validate_response(response)
         return if response.code == 200
 
+        if response.headers['content-type'].start_with?('text/html')
```
d-Pixie, line 19: This also commonly happen when their API is down etc. For some reason you get a default 500 HTML page as response when they are down ...

spec/support/helpers/configuration_helper.rb
```diff
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 
+require 'dotenv'
+
+Dotenv.load('.env.test')
```
d-Pixie, line 5: Are you sure this is needed? I read the docs and it look like it would pick this up automatically if it exists.

README.md
```diff
@@ -104,47 +104,127 @@ $ gem install fortnox-api
...
+  require 'redis'
+
+  attr_accessor :__access_token
+  attr_accessor :__refresh_token
+
+  alias_method :access_token, :__access_token
+  alias_method :refresh_token, :__refresh_token
+
+  def initialize
+    // TODO: Initialize redis...
```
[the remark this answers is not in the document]
ehannes: Nice!

README.md
```diff
@@ -104,47 +104,127 @@ $ gem install fortnox-api
...
+  end
+
+  def refresh_token=(token)
+    __refresh_token = token
+    redis.set(refresh_token_key, token)
+  end
+
+  private
+    def access_token_key
+      "#{@prefix}access_token"
```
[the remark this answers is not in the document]
ehannes: Much better, thanks!

README.md
```diff
@@ -104,47 +104,127 @@ $ gem install fortnox-api
...
+  attr_accessor :__access_token
+  attr_accessor :__refresh_token
+
+  alias_method :access_token, :__access_token
+  alias_method :refresh_token, :__refresh_token
+
+  def initialize(account: nil)
+    // TODO: Initialize redis...
+
+    @prefix = account.empty? ? '' : "#{account}_"
```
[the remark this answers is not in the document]
ehannes: Much better, thanks!

bin/get_tokens
ehannes: Do you mean "Constant?" as in "is this is always the endpoint"? In that case - yes, this is the endpoint Fortnox uses according to https://developer.fortnox.se/general/authentication/.

spec/support/helpers/configuration_helper.rb
ehannes: Correct. I did need it in my `bin/get_tokens`, but not here.

lib/fortnox/api/repositories/base.rb
ehannes: This is a good idea, but you will get `undefined method `[]' for nil:NilClass`

spec/support/helpers/configuration_helper.rb
ehannes: Hmm, actually, I can't get it to work without explicitly loading `.env.test`...

bin/get_tokens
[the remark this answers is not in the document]
ehannes: I've changed this script now. Please have a second look at it.

lib/fortnox/api.rb
```diff
@@ -26,22 +26,11 @@ module API
...
-      value
-    end
-    setting :access_tokens, DEFAULT_CONFIGURATION[:access_tokens] do |value|
-      next if value.nil? # nil is a valid unassigned value
-
-      invalid_access_tokens_format!(value) unless value.is_a?(Hash) || value.is_a?(Array)
-      config.token_store = value.is_a?(Hash) ? value : { default: value }
-      value
-    end
+    setting :storage, DEFAULT_CONFIGURATION[:storage]
```
d-Pixie, line 31: I also realised that we should just alias `token_store = store` to be == `token_stores[:default] = store`, that way we can treat the stores the same throughout our code ...

lib/fortnox/api/repositories/base.rb
```diff
@@ -100,11 +103,90 @@ def config
...
 
-        def token_store_not_found!(store_name)
-          raise MissingConfiguration,
-                "There is no token store named #{store_name}. Available stores are #{config.token_store.keys}."
+        def get_token_store(name)
+          storage = config.storages&.dig(name.to_sym) || config.storage
+          validate_store(storage, name)
+        end
+
+        def validate_store(store, name)
```
d-Pixie, line 111: I'd say no ... This is defensive programming at its worst. The store will throw if we try to access a `name` that it doesn't have. If you want to give a nicer message you can catch that error and reformat it there.

We have given an API that the developers of stores have to adhere to, if you want more control you should make an interface that they have to implement, something like:
```
class Fortnox::API::TokenStore
  class MissingImplementation < StandardError; end
  def access_token(name = :default)
    raise MissingImplementation, "TokenStore must implement access_token="
  end
  ...
end
```

lib/fortnox/api/repositories/base.rb
```diff
@@ -100,11 +103,90 @@ def config
...
+
+          unless store.respond_to? :refresh_token=
+            raise MissingConfiguration,
+                  "Store named #{name} does not implement required #refresh_token="
+          end
+
+          store
+        end
+
+        def time_to_renew?(token)
```
d-Pixie, line 141: I would rename to `expired?`.

d-Pixie (changes requested):

Rakefile
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/MethodLength
+
 require 'rspec/core/rake_task'
+require_relative 'lib/fortnox/api'
```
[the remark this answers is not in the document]
ehannes: Agreed!

bin/get_tokens
```diff
@@ -0,0 +1,84 @@
...
+require 'dotenv'
+require 'httparty'
+
+dotenv_file_name = '.env'
+
+Dotenv.load(dotenv_file_name)
+
+print "Let's get you an access token and a refresh token for Fortnox API."
+
+client_id = ENV.fetch('FORTNOX_API_CLIENT_ID')
```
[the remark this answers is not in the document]
ehannes: Sure, much nicer!

.env.test
```diff
@@ -1,3 +1 @@
-FORTNOX_API_BASE_URL=https://api.fortnox.se/3/
-FORTNOX_API_CLIENT_SECRET=9aBA8ZgsvR
-FORTNOX_API_ACCESS_TOKEN=ccaef817-d5d8-4b1c-a316-54f3e55c5c54
+MOCK_VALID_ACCESS_TOKEN=true
```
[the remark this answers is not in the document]
ehannes: Ye, that's even better. Good idea!

lib/fortnox/api/repositories/base.rb
```diff
@@ -46,32 +54,20 @@ def self.set_headers(headers = {}) # rubocop:disable Naming/AccessorMethodName
...
-
-        def check_access_tokens!(tokens)
-          tokens_present = !(tokens.nil? || tokens.empty?)
-          return if tokens_present
-
-          error_message = "You have not provided any access tokens in token store #{@token_store.inspect}."
-          raise MissingConfiguration, error_message
-        end
+        def access_token
+          token_from_store = @token_store.access_token
```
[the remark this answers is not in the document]
ehannes: Agreed. I refactored the code and now `token` is available :+1:

lib/fortnox/api/repositories/base.rb
```diff
@@ -89,6 +85,13 @@ def base_url
           base_url
         end
 
+        def client_id
+          client_id = config.client_id
```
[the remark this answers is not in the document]
ehannes: Where do we check this you mean? I though this was the place where the check is taking place. There are no constructors left in `lib/fortnox/api.rb`. Maybe I removed them by mistake, or am I missing something else here? :thinking:

.env.test
```diff
@@ -1 +1,4 @@
-MOCK_VALID_ACCESS_TOKEN=true
+# Fake long lived token, generated with:
+# JWT.encode({ exp: 1_906_691_557 }, nil, 'none')
+# Expires 1906691557 = Mon Jun 03 2030 04:32:37 GMT+0000
+FORTNOX_API_ACCESS_TOKEN=eyJhbGciOiJub25lIn0.eyJleHAiOjE5MDY2OTE1NTd9.
```
d-Pixie, line 4: 👍

README.md
```diff
@@ -214,7 +216,7 @@ class MyStorage
 end
 
 Fortnox::API.configure do |config|
-  config.storages = {
+  config.token_stores = {
     default: MyStorage.new,
     another_account: MyStorage.new(account: :another_account)
```
d-Pixie, line 221: Why is the `(account: :anothet_account)` needed? You are already keeping them under keys in the token_store hash. Each one is an independent instance of a store so ...

bin/get_tokens
```diff
@@ -5,35 +5,30 @@ require 'base64'
 require 'dotenv'
 require 'httparty'
 
-dotenv_file_name = '.env'
+DOTENV_FILE_NAME = '.env'
```
d-Pixie, line 8: Thanks 🙂

bin/get_tokens
```diff
@@ -5,35 +5,30 @@ require 'base64'
...
 
-print "Let's get you an access token and a refresh token for Fortnox API."
-
-client_id = ENV.fetch('FORTNOX_API_CLIENT_ID')
-client_secret = ENV.fetch('FORTNOX_API_CLIENT_SECRET')
-redirect_uri = ENV.fetch('FORTNOX_API_REDIRECT_URI')
-scopes = ENV.fetch('FORTNOX_API_SCOPES')
-credentials = Base64.encode64("#{client_id}:#{client_secret}").to_s
-oauth_endpoint = 'https://apps.fortnox.se/oauth-v1'
+CLIENT_ID = ENV.fetch('FORTNOX_API_CLIENT_ID'){ |key| raise "Missing environment variabel #{key}! Exiting." }
```
d-Pixie, line 13: Is this allowed? I would think that this would cause a "Dynamic assignment to constant" error ...

bin/get_tokens
```diff
@@ -5,35 +5,30 @@ require 'base64'
...
+CLIENT_SECRET = ENV.fetch('FORTNOX_API_CLIENT_SECRET'){ |key| raise "Missing environment variabel #{key}! Exiting." }
+REDIRECT_URI = ENV.fetch('FORTNOX_API_REDIRECT_URI'){ |key| raise "Missing environment variabel #{key}! Exiting." }
+SCOPES = ENV.fetch('FORTNOX_API_SCOPES'){ |key| raise "Missing environment variabel #{key}! Exiting." }
```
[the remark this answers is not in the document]
d-Pixie: Feel free to ignore ...

bin/get_tokens
```diff
@@ -5,35 +5,30 @@ require 'base64'
...
 
-raise 'Missing client ID! Exiting.' if client_id.empty?
-raise 'Missing client secret! Exiting' if client_secret.empty?
-raise 'Missing redirect_uri! Exiting.' if redirect_uri.empty?
-raise 'Missing scopes! Exiting.' if scopes.empty?
+credentials = Base64.encode64("#{CLIENT_ID}:#{CLIENT_SECRET}").to_s
 
 auth_params =
```
d-Pixie, line 20: This could equally be a constant then, right?

lib/fortnox/api/repositories/base.rb
```diff
@@ -55,19 +55,21 @@ def self.set_headers(headers = {}) # rubocop:disable Naming/AccessorMethodName
         def initialize(keys_filtered_on_save: [:url], token_store: :default)
           @keys_filtered_on_save = keys_filtered_on_save
           @token_store_name = token_store
-          @token_store = get_token_store(token_store)
+          @token_store = config.token_stores&.dig(token_store.to_sym)
           @mapper = Registry[Mapper::Base.canonical_name_sym(self.class::MODEL)].new
+
+          validate_store(@token_store, @token_store_name)
```
d-Pixie, line 61: I would suggest changing this to:
```
raise( MissingConfiguration, "There is no token store named \"#{token_store}\". " \"config.token_stores: #{config.token_stores}.") if @token_store.nil?

File: lib/fortnox/api/repositories/base.rb [hunk 31]
```diff
@@ -55,19 +55,21 @@ def self.set_headers(headers = {}) # rubocop:disable Naming/AccessorMethodName
         def initialize(keys_filtered_on_save: [:url], token_store: :default)
           @keys_filtered_on_save = keys_filtered_on_save
           @token_store_name = token_store
-          @token_store = get_token_store(token_store)
+          @token_store = config.token_stores&.dig(token_store.to_sym)
```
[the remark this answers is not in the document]
ehannes: Ah, yes, it's a leftover from the previous implementation, thanks! :+1:

README.md
ehannes: It's for the Redis part in the token store: we set `@prefix = account.empty? ? :default : "#{account}"` and `access_token_key` is then `"#{@prefix}_access_token"`. If we don't pass the `account` parameter to the second instance of the token store, the stores will overwrite each others keys since the Redis instance is assumed to be shared between the token stores.

ehannes: I will rename `MyStorage` to `MyTokenStore` though since we have renamed `storages` to `token_stores`.

bin/get_tokens
ehannes: Yes, it works just as intended.

bin/get_tokens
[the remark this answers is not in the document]
ehannes: Wow, that's really nice. I thought about that but didn't know about this feature in Ruby, thanks for teaching me!

lib/fortnox/api/repositories/base.rb
```diff
@@ -55,19 +55,21 @@ def self.set_headers(headers = {}) # rubocop:disable Naming/AccessorMethodName
         def initialize(keys_filtered_on_save: [:url], token_store: :default)
           @keys_filtered_on_save = keys_filtered_on_save
           @token_store_name = token_store
```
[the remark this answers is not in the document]
ehannes: It's used for better error messages here:
```
        def expired?(token)
          decoded_token = JWT.decode token, nil, false
          decoded_token[0]['exp'] < (Time.now.to_i - TIME_MARGIN_FOR_ACCESS_TOKEN_RENEWAL)
        rescue JWT::DecodeError
          raise Exception, "Could not decode access token for token store \"#{@token_store_name}\""
        end

```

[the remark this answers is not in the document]
d-Pixie: Well, since you set that in the token store instance, for use as a Redis key, there should be no reason to remember it here as well. Expose it on the TokenStore and just do: `raise Exception, "Could not decode access token for token store \"#{@token_store.name}\""`?

PR #221, opened 2022-06-02
Changes to relative_require

ehannes:
Changes all import inside the gem to `require_relative` to fix the problem with loading the current development version of the gem in Rake tasks.

I'm no big fan of the `.../` that we have now. Not sure if we can do anything about that...


lib/fortnox/api/models/customer.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require 'fortnox/api/types'
-require 'fortnox/api/models/base'
+require_relative 'base'
+require_relative '..//types'
```
d-Pixie, line 4: Double slash

d-Pixie (changes requested):

PR #226, opened 2023-01-12
Fix broken dependencies

### LINKED ISSUE #209: Fix deprecation warnings from dry-configurable
```
~/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/dry-types-0.12.3/lib/dry/types.rb:24:in `<module:Dry>' [dry-configurable] default value as positional argument to settings is depr
ecated and will be removed in the next major version                                                                                                                                          
Provide a `default:` keyword argument instead
```

ehannes:
- Fixes #209 
- Adds missing dependencies which broke the gem
- Drops support for Ruby 2.6 (dry-configurable 0.15 requires 2.7)

Thanks to @maokomioko, I was able to use some of your code to do these changes. It's not our intention to ignore your PR but as we said, we need to fix this gem now and your PR is mixing a lot of changes which makes it really hard to reuse. Please contact us so that we can have a talk about it 😄 

I will continue upgrading the dependencies in upcoming pull requests the coming days.


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.7.6
```
ehannes, line 1: I think it's better to ship a version of Ruby with this gem that we as developers use, so that we know we are using the same version. Much easier to debug things when you know everyone is using the same version...

fortnox-api.gemspec
```diff
@@ -24,8 +24,10 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
 
   spec.required_ruby_version = '>= 2.6'
   spec.add_dependency 'countries', '~> 3.0'
-  spec.add_dependency 'dry-struct', '~> 0.1'
-  spec.add_dependency 'dry-types', '~> 0.8', '< 0.13.0'
+  spec.add_dependency 'dry-configurable', '0.15.0'
```
ehannes, line 27: This was missing, which lead to broken dependencies.

fortnox-api.gemspec
```diff
@@ -24,8 +24,10 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
...
+  spec.add_dependency 'dry-container', '0.9.0'
```
ehannes, line 28: This was missing, which lead to broken dependencies.

fortnox-api.gemspec
```diff
@@ -24,8 +24,10 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
...
+  spec.add_dependency 'dry-struct', '0.4.0'
+  spec.add_dependency 'dry-types', '0.12.3'
```
ehannes, line 30: 0.13.0 is breaking. I'll update that in a later PR...

fortnox-api.gemspec
```diff
@@ -24,8 +24,10 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
 
   spec.required_ruby_version = '>= 2.6'
   spec.add_dependency 'countries', '~> 3.0'
-  spec.add_dependency 'dry-struct', '~> 0.1'
-  spec.add_dependency 'dry-types', '~> 0.8', '< 0.13.0'
+  spec.add_dependency 'dry-configurable', '0.15.0'
+  spec.add_dependency 'dry-container', '0.9.0'
+  spec.add_dependency 'dry-struct', '0.4.0'
```
ehannes, line 29: This was missing, which lead to broken dependencies. Let's go with fixed versions for now so that we can upgrade this one step at a time.

lib/fortnox/api.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 require 'set'
-require 'dry-struct'
```
ehannes, line 4: Not sure why we required `dry-struct` here before. We are not using that in this file. What we are using is `dry-container` and `dry-configurable`...

lib/fortnox/api.rb
```diff
@@ -24,13 +25,13 @@ module API
...
     }.freeze
 
-    setting :base_url, DEFAULT_CONFIGURATION[:base_url]
-    setting :client_id, DEFAULT_CONFIGURATION[:client_id]
-    setting :client_secret, DEFAULT_CONFIGURATION[:client_secret]
-    setting :token_stores, DEFAULT_CONFIGURATION[:token_stores]
-    setting :token_url, DEFAULT_CONFIGURATION[:token_url]
-    setting :debugging, DEFAULT_CONFIGURATION[:debugging], reader: true
-    setting :logger, DEFAULT_CONFIGURATION[:logger], reader: true
+    setting :base_url, default: DEFAULT_CONFIGURATION[:base_url]
```
ehannes, line 28: Fixes deprecation warning.

ehannes:
Build for Ruby 3 is still broken, which is expected...

CHANGELOG.md
```diff
@@ -1,58 +1,88 @@
...
-  that Fortnox has implemented. The configuration part of the gem is new, have a look at the readme.
+
+- **Breaking** The auth process is rewritten to support
+  [the new auth flow](https://developer.fortnox.se/general/authentication/) that
+  Fortnox has implemented. The configuration part of the gem is new, have a look
+  at the readme.
+
+### Fixed
+
+- **Breaking** Adding missing runtime dependencies dry-configurable and
```
d-Pixie, line 24: This is not breaking.

fortnox-api.gemspec
d-Pixie: Fine since that dependency is at a 0.x version.

lib/fortnox/api.rb
d-Pixie: Probably a leftover from refactoring ...

d-Pixie (approved):

ehannes:
Closed in favor of #228

Issue #212, opened 2022-04-28
Add support for Ruby 3


ehannes:
Travis build is not even working now:

```
Setting environment variables from .travis.yml
$ export CC_TEST_REPORTER_ID=[filtered-id]
1.10s$ rvm get head
Downloading https://get.rvm.io
Downloading https://bitbucket.org/mpapis/rvm/raw/master/binscripts/rvm-installer.asc
Micros.asc
The command "rvm get head" failed and exited with 3 during .
Your build has been stopped.
```

ehannes:
e93fb785108497dede6248928105b64f9ea2c4ef fixed the above error. Simply added an explicit Ubuntu version to run Travis with. It was by default running on Xenial before, which is Ubuntu 16.04! That's 5 years old now, time to upgrade! Interesting that Travis uses that distro by default...

Now we have another problem, in both 3.0.4 and 3.1.2:
```
Failure/Error: require 'dry-struct'
ArgumentError:
  wrong number of arguments (given 2, expected 1)
```

ehannes:
The above problem might be because the current used version of `dry-struct` is not supporting Ruby 3. That version is released 2017-11-04, more than 4 years ago and long before Ruby 3 was around!

ehannes:
Resolved in #228

PR #228, opened 2023-01-13
Dry libs upgrade

### LINKED ISSUE #177: Pinned dry-types version
`dry-types` is pinned to version < 0.13.0 due to breaking changes. See https://github.com/accodeing/fortnox-api/commit/1f446ef00dbd15fe7df1c33d997c394a8b883063. Unfortunately, I have no idea what the breaking changes is. But probably something we should fix sometime...

### LINKED ISSUE #209: Fix deprecation warnings from dry-configurable

ehannes:
- Fixes #209 
- Adds missing dependencies which broke the gem
- Drops support for Ruby 2.6
- Fixes broken Travis builds and therefor also #212 
- Fixes #177 

Many thanks to @maokomioko for all the work!


ehannes:
Travis build finally successful! 😄

.rubocop.yml
```diff
@@ -1,7 +1,7 @@
 require: rubocop-rspec
 
 AllCops:
-  TargetRubyVersion: 2.6
+  TargetRubyVersion: 2.7
```
d-Pixie, line 4: Shouldn't this be a 3.x version if that is what we are targeting?

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.7.7
```
d-Pixie, line 1: As should this ...

fortnox-api.gemspec
```diff
@@ -22,25 +22,27 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
...
+  spec.add_dependency 'dry-struct', '~> 1.6.0'
+  spec.add_dependency 'dry-types', '~> 1.7.0'
+  spec.add_dependency 'httparty', '~> 0.17.3'
+  spec.add_dependency 'jwt', '~> 2.6.0'
 
-  spec.add_development_dependency 'bundler', '~> 2.0'
-  spec.add_development_dependency 'dotenv', '~> 2.7'
-  spec.add_development_dependency 'guard', '~> 2.12'
-  spec.add_development_dependency 'guard-rspec', '~> 4.5'
+  spec.add_development_dependency 'bundler', '>= 2.4'
```
d-Pixie, line 34: Why `>=` here and not `~>`?

fortnox-api.gemspec
```diff
@@ -22,25 +22,27 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
...
+  spec.add_development_dependency 'dotenv', '~> 2.8'
+  spec.add_development_dependency 'guard', '~> 2.18'
+  spec.add_development_dependency 'guard-rspec', '~> 4.7'
   spec.add_development_dependency 'pry', '~> 0'
-  spec.add_development_dependency 'rake', '>= 12.3.3'
-  spec.add_development_dependency 'rspec', '~> 3.2'
+  spec.add_development_dependency 'rake', '>= 13.0.6'
```
d-Pixie, line 39: And here...

lib/fortnox/api/models/base.rb
```diff
@@ -87,17 +76,9 @@ def parent
...
-
-          self.class.schema.keys.each_with_object({}) do |key, result|
-            result[key] = self[key]
-          end
-        end
-
         # dry-types filter anything that isn't specified as an attribute on the
         # class that is being instantiated. This wrapper preserves the meta
-        # properties we need to track object state during that initilisation and
+        # properties we need to track object state during that initialization and
```
d-Pixie, line 81: Should be `initialisation`, with an `s`, since we/I standardise on British english...

lib/fortnox/api/types/nullable.rb
```diff
@@ -8,13 +8,17 @@ module Types
...
-        Boolean = Types::Bool.optional.constructor { |value| THE_TRUTH.fetch(value) { false } }
-        # TODO: Improve parsing!
-        # In case date parsing fails, ArgumentError is thrown. Currently, it is rescued in Repository::Loaders.find.
-        # That method assumes that the exception is due to invalid argument to the find method, which is not the case
-        # if it is raised from here!
-        Date = Types::Date.optional.constructor { |value| ::Date.parse(value.to_s) unless value.nil? || value == '' }
+        Integer = Types::Strict::Integer.optional.constructor { |value| value&.to_i }
+        Boolean = Types::Bool.optional.constructor { |value| THE_TRUTH.fetch(value, false) }
+        Date = Types::Date.optional.constructor do |value|
+          next if value.nil? || value == ''
```
d-Pixie, line 14: I don't think `next` is correct here. We probably want to do `return nil` instead, to avoid letting `''` through as a nil value. Or does that get handled internally by Dry::Types?

lib/fortnox/api/types/order_row.rb
```diff
@@ -9,7 +9,7 @@ class OrderRow < DocumentRow
         STUB = { ordered_quantity: 0 }.freeze
 
         # OrderedQuantity Ordered quantity
-        attribute :ordered_quantity, Types::Required::Float
+        attribute :ordered_quantity, ::Fortnox::API::Types::Required::Float
```
d-Pixie, line 12: Is the prefix needed? Shouldn't be ...

spec/fortnox/api/mappers/default_delivery_types_spec.rb
```diff
@@ -8,7 +8,7 @@
 describe Fortnox::API::Mapper::DefaultDeliveryTypes do
   key_map = {}
 
-  it_behaves_like 'mapper', key_map do
+  it_behaves_like 'mapper', key_map, nil, nil do
```
d-Pixie, line 11: Arn't these already defaulted to `nil`?

spec/support/matchers/type/have_nullable_string_matcher.rb
```diff
@@ -33,13 +33,13 @@ def accepts_string?
       def rejects_non_string?
         non_string = 10.0
         @klass.new(@valid_hash.merge(@attribute => non_string))
-      rescue Fortnox::API::InvalidAttributeValueError => error
+      rescue Fortnox::API::AttributeError => e
```
d-Pixie, line 36: It's doing that thing again. We should change this cop ...

d-Pixie (changes requested):

.rubocop.yml
ehannes: Correct!

.tool-versions
ehannes: Yep

.rubocop.yml
ehannes: Actually, we were both wrong. This should be the oldest supported version of Ruby.

spec/support/matchers/type/have_nullable_string_matcher.rb
ehannes: I'll fix this in an upcoming Rubocop branch...

lib/fortnox/api/types/order_row.rb
ehannes: I removed it and it still worked, so let's stick with just `Types::...` 😄 Don't know why this was changed in the first place.

lib/fortnox/api/types/nullable.rb
ehannes: We can't use `return` in a `constructor`... That's why we are using `next`.

ehannes: It's handled by `Dry::Types`. I added a test for it and it works as expected. `''` gives `nil`.

PR #229, opened 2023-01-13
Rubocop fixes

### LINKED ISSUE #178: Improve container testing
We can remove a hack in our test according to https://github.com/dry-rb/dry-container/issues/62

ehannes:
When this is merged, there's another branch that should be reviewed with even more fixes 😄 

Fixes #178


spec/fortnox/api/mappers/contexts/json_conversion.rb
```diff
@@ -2,53 +2,66 @@
...
+    stub_const('Test::ProductDesignerMapper', Class.new(Test::BaseMapper))
+    stub_const('Test::ProductDesignerMapper::KEY_MAP', { id: 'ID' }.freeze)
 
-      class Category < Fortnox::API::Model::Base
+    stub_const('Test::ProductMapper', Class.new(Test::BaseMapper))
+    stub_const(
+      'Test::ProductMapper::KEY_MAP',
+      {
+        vat: 'VAT',
+        url: '@url' # TODO: How to handle url attribute?
```
ehannes, line 24: Old comment...

spec/fortnox/api/models/base_spec.rb
```diff
@@ -24,35 +24,33 @@
     end
 
     context 'without required attribute' do
-      subject { -> { entity_class.new({}) } }
-
-      it { is_expected.to raise_error Fortnox::API::Exception }
```
ehannes, line 29: Removes some unnecessary tests here...

spec/fortnox/api/types/housework_types_spec.rb
```diff
@@ -42,49 +45,46 @@
...
+  it_behaves_like 'housework type', 'GARDENING', 'TYPE_RUT'
+  it_behaves_like 'housework type', 'BABYSITTING', 'TYPE_RUT'
+  it_behaves_like 'housework type', 'OTHERCARE', 'TYPE_RUT'
+  it_behaves_like 'housework type', 'OTHERCOSTS', 'TYPE_RUT', housework: false
 
   # rubocop:disable RSpec/RepeatedExample
   # rubocop:disable RSpec/RepeatedDescription
   pending 'to be added' do
-    raise Exception, 'Will be supported 2021-01-01'
+    raise StandardError, 'Will be supported 2021-01-01'
```
ehannes, line 79: Rubocop suggested changing from Exception to StandardError

spec/spec_helper.rb
```diff
@@ -26,19 +26,6 @@
 
   WebMock.disable_net_connect!(allow: 'codeclimate.com')
 
-  config.before do
```
ehannes, line 29: Not needed anymore since we are using stubs instead

.rubocop.yml
```diff
@@ -23,3 +23,15 @@ Metrics/BlockLength:
 Naming/PredicateName:
   Exclude:
     - 'spec/support/matchers/**/*'
+
+Naming/RescuedExceptionsVariableName:
+  PreferredName: exception
```
d-Pixie, line 28: ❤️

.rubocop.yml
```diff
@@ -23,3 +23,15 @@ Metrics/BlockLength:
...
+
+Style/WordArray:
+  EnforcedStyle: brackets
```
d-Pixie, line 31: ❤️

lib/fortnox/api/repositories/base.rb
```diff
@@ -111,7 +111,7 @@ def expired?(token)
           decoded_token = JWT.decode token, nil, false
           decoded_token[0]['exp'] < (Time.now.to_i - TIME_MARGIN_FOR_ACCESS_TOKEN_RENEWAL)
         rescue JWT::DecodeError
-          raise Exception, "Could not decode access token for token store \"#{@token_store_name}\""
+          raise Fortnox::API::Exception, "Could not decode access token for token store \"#{@token_store_name}\""
```
d-Pixie, line 114: Why do you need this if you have:
```
Lint/RaiseException:
  AllowedImplicitNamespaces: ['Fortnox::API']
```

lib/fortnox/api/types/enums.rb
```diff
@@ -36,16 +36,11 @@ def self.lower_case
...
-          MAJORAPPLIANCEREPAIR MOVINGSERVICES ITSERVICES CLEANING
-          TEXTILECLOTHING SNOWPLOWING GARDENING BABYSITTING OTHERCARE
-          OTHERCOSTS
-        ],
-        legacy_rut: %w[COOKING TUTORING]
+        rot: ['CONSTRUCTION', 'ELECTRICITY', 'GLASSMETALWORK', 'GROUNDDRAINAGEWORK', 'MASONRY', 'PAINTINGWALLPAPERING',
+              'HVAC', 'OTHERCOSTS'],
+        rut: ['MAJORAPPLIANCEREPAIR', 'MOVINGSERVICES', 'ITSERVICES', 'CLEANING', 'TEXTILECLOTHING', 'SNOWPLOWING',
+              'GARDENING', 'BABYSITTING', 'OTHERCARE', 'OTHERCOSTS'],
+        legacy_rut: ['COOKING', 'TUTORING']
```
d-Pixie, line 43:
```suggestion
        rot: [
          'CONSTRUCTION',
          'ELECTRICITY',
          'GLASSMETALWORK',
          'GROUNDDRAINAGEWORK',
          'MASONRY',
          'PAINTINGWALLPAPERING',
          'HVAC',
          'OTHERCOSTS'
        ],
        rut: [
          'MAJORAPPLIANCEREPAIR',
          'MOVINGSERVICES',
          'ITSERVICES',
          'CLEANING',
          'TEXTILECLOTHING',
          'SNOWPLOWING',
          'GARDENING',
          'BABYSITTING',
          'OTHERCARE',
          'OTHERCOSTS'
        ],
        legacy_rut: [
          'COOKING',
          'TUTORING'
        ]
```

spec/fortnox/api/repositories/base_spec.rb
```diff
@@ -221,19 +221,19 @@ def refresh_token=; end
...
+                         ''
+                       end
 
-              def access_token=; end
+                       def access_token=; end
 
-              def refresh_token=; end
-            end.new
+                       def refresh_token=; end
+                     end.new
```
d-Pixie, line 236: Why not just make it a module and skip the need for `.new`?

spec/fortnox/api/types/housework_types_spec.rb
```diff
@@ -25,7 +28,7 @@
     let(:document) do
       Fortnox::API::Model::Order.new(
         customer_number: '1',
-        tax_reduction_type: tax_reduction_type,
+        tax_reduction_type: Object.const_get(tax_reduction_type),
```
d-Pixie, line 31: Hmmm, this looks weird ...

spec/fortnox/api/types/housework_types_spec.rb
```diff
@@ -42,49 +45,46 @@
         let(:error_message) { 'Skattereduktion för den valda typen av husarbete har upphört.' }
 
         it 'raises an error' do
```
d-Pixie, line 47: This could quite easily be rewritten to
```
it is_expected.to raise_error(Fortnox::API::RemoteServerError, error_message)
```
since you don't really need the description there ...

spec/fortnox/api/types/housework_types_spec.rb
d-Pixie: Do we need this anymore? The "Will be supported" date has passed ...

spec/fortnox/api/types/housework_types_spec.rb
```diff
@@ -42,49 +45,46 @@
...
+    raise StandardError, 'Will be supported 2021-01-01'
     # it_behaves_like 'housework type', 'HOMEMAINTENANCE', TYPE_RUT
     # it_behaves_like 'housework type', 'FURNISHING', TYPE_RUT
     # it_behaves_like 'housework type', 'TRANSPORTATIONSERVICES', TYPE_RUT
     # it_behaves_like 'housework type', 'WASHINGANDCAREOFCLOTHING', TYPE_RUT
   end
 
   pending 'to be added' do
-    raise Exception, 'Will be supported 2021-01-01'
+    raise StandardError, 'Will be supported 2021-01-01'
```
d-Pixie, line 87: As has this

spec/fortnox/api/types/model_spec.rb
```diff
@@ -42,7 +38,7 @@
       stub_const('User', user_class)
     end
 
-    it { is_expected.not_to raise_error }
+    it { expect { User.new }.not_to raise_error }
```
d-Pixie, line 41: Don't love how this reads ...

spec/fortnox/api/types/required_spec.rb
```diff
@@ -15,25 +15,27 @@
     end
 
     it 'raises an error' do
-      is_expected.to raise_error(Dry::Struct::Error, error_message)
+      expect(subject).to raise_error(Dry::Struct::Error, error_message)
```
d-Pixie, line 18: Same here about dropping the description and keeping the is_expected form.

d-Pixie: Also, if this can still be on the form where `subject` is expected to raise an error, why couldn't the last one ^?

spec/fortnox/api/types/housework_types_spec.rb
ehannes: Does it? I just pass the class as a string and use `Object.const_get` to "constantize" the string. We don't have the `constantize` method from Rails to use here...

spec/fortnox/api/types/housework_types_spec.rb
ehannes: Ye, I need to implement it first 😅 Will do in a follow up PR...

spec/fortnox/api/types/housework_types_spec.rb
ehannes: Same here

spec/fortnox/api/types/model_spec.rb
ehannes: Ye, I agree. Problem is that Rubocop doesn't like when you have a proc as the `subject`. Not sure why though... I thought it was regarded as bad practice, but maybe this is something we should discuss internally.

lib/fortnox/api/repositories/base.rb
ehannes: We don't. Good catch 👍

ehannes: Unfortunately, I can't get this to work. I have this config:
```
Lint/RaiseException:
  AllowedImplicitNamespaces: ['Fortnox::API']
```
but I still get this error:
```
lib/fortnox/api/repositories/base.rb:114:17: W: [Correctable] Lint/RaiseException: Use StandardError over Exception.
          raise Exception, "Could not decode access token for token store \"#{@token_store_name}\""
                ^^^^^^^^^
```

Maybe Rubocop is not smart enough. In the [documentation](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/RaiseException) they have an example of `Gem::Exception`, but that exception is directly in a module named `Gem`. Here, we have a class named `Fortnox::API::Repository::Base` that uses the `Fortnox::API::Exception` exception. Don't know how to fix it. I will use the fully qualified name for now...

spec/fortnox/api/types/housework_types_spec.rb
ehannes: Before, I created variables for `TYPE_ROT` and `TYPE_RUT`. Now I stub them, which I need to do in an RSpec context, which means they are no longer declared when I run `it_behaves_like 'housework type', 'CONSTRUCTION', 'TYPE_ROT'` (`'TYPE_ROT'` is now a string, before it was a class).

spec/fortnox/api/types/model_spec.rb
ehannes: This is where we should use `specify` instead of `it`...

spec/fortnox/api/types/required_spec.rb
ehannes: We should not have subject raising an error. See https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html

spec/fortnox/api/repositories/base_spec.rb
ehannes: Then we would need to stub it to not leak the module between tests. Better to just have it as an anonymous class as it is...

PR #232, opened 2023-01-18
Dry libs upgrade rubocop fixes new cops

ehannes:
More Rubocop fixes. Adds **a lot** of new Rubocop cops.

### DIFF
--- a/fortnox-api.gemspec
+++ b/fortnox-api.gemspec
@@ -11,9 +11,9 @@ Gem::Specification.new do |spec|
   spec.email         = ['info@accodeing.com']
   spec.summary       = 'Gem to use Fortnox REST API in Ruby.'
   spec.description   = 'This gem uses the HTTParty library to abstract away the REST calls. It gives you access to a ' \
-                       'number of objects that behave a lot like ActiveRecord instances, giving you access to methods ' \
-                       'like `all`, `find`, `find_by_...` and so on. And each individual instance can be easily' \
-                       'persistable to Fortnox again using the `save` method.'
+                       'number of objects that behave a lot like ActiveRecord instances, giving you access to ' \
+                       'methods like `all`, `find`, `find_by_...` and so on. And each individual instance can be' \
+                       'easily persistable to Fortnox again using the `save` method.'
   spec.homepage      = 'http://github.com/accodeing/fortnox-api'
   spec.license       = 'LGPL-3.0'
 
--- a/fortnox-api.gemspec
+++ b/fortnox-api.gemspec
@@ -16,6 +16,9 @@ Gem::Specification.new do |spec|
                        'easily persistable to Fortnox again using the `save` method.'
   spec.homepage      = 'http://github.com/accodeing/fortnox-api'
   spec.license       = 'LGPL-3.0'
+  spec.metadata = {
+    'rubygems_mfa_required' => 'true'
+  }
 
   spec.files         = `git ls-files -z`.split("\x0")
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
@@ -45,5 +48,4 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'timecop', '~> 0.9.6'
   spec.add_development_dependency 'vcr', '~> 6.1'
   spec.add_development_dependency 'webmock', '~> 3.18'
-  spec.metadata['rubygems_mfa_required'] = 'true'
 end
--- a/fortnox-api.gemspec
+++ b/fortnox-api.gemspec
@@ -10,16 +10,15 @@ Gem::Specification.new do |spec|
   spec.authors       = ['Jonas Schubert Erlandsson', 'Hannes Elvemyr', 'Felix Holmgren', 'Mike Eirih']
   spec.email         = ['info@accodeing.com']
   spec.summary       = 'Gem to use Fortnox REST API in Ruby.'
-  spec.description   = 'This gem uses the HTTParty library to abstract away the REST calls. It gives you access to a '\
-                       'number of objects that behave a lot like ActiveRecord instances, giving you access to methods '\
-                       'like `all`, `find`, `find_by_...` and so on. And each individual instance can be easily'\
+  spec.description   = 'This gem uses the HTTParty library to abstract away the REST calls. It gives you access to a ' \
+                       'number of objects that behave a lot like ActiveRecord instances, giving you access to methods ' \
+                       'like `all`, `find`, `find_by_...` and so on. And each individual instance can be easily' \
                        'persistable to Fortnox again using the `save` method.'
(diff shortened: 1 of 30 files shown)


ehannes:
@d-Pixie what do you think about the code duplication that codeclimate is complaining about? Looks like something we should not bother about... It's just similar test setup...

d-Pixie:
> 

I say ignore it. I agree with you that extracting that wouldn't make the code better, but it would make it harder to read so ... Easy choice 🙂

lib/fortnox/api/types/model.rb
```diff
@@ -18,7 +18,7 @@ def initialize(input_attributes)
         private
 
         def missing_keys(attributes)
-          non_nil_attributes = attributes.reject { |_, value| value.nil? }
+          non_nil_attributes = attributes.compact
```
d-Pixie, line 21: I'd be for folding this into the next line now, since it is such a simple to read line `attribute_keys = attributes.compact.keys`

spec/fortnox/api/mappers/base/from_json_spec.rb
```diff
@@ -24,10 +24,10 @@
       {
         'Product' => {
           '@url': 'someurl@example.com',
-          'Name': 'Ford Mustang',
-          'VAT': 30_000,
-          'Categories': [{ 'Name' => 'Cars', 'ID' => '1' }, { 'Name' => 'Fast cars', 'ID' => '2' }],
-          'Designer': { 'Name' => 'John Najjar', 'ID' => '23' }
+          Name: 'Ford Mustang',
```
d-Pixie, line 27: Ugh ... I know what it's doing, but I do like it when the same convention is used in every case - at least locally 😄 (It's fine ... I just wanted to complain a bit)

spec/fortnox/api/mappers/base/from_json_spec.rb
```diff
@@ -24,10 +24,10 @@
...
+          VAT: 30_000,
+          Categories: [{ 'Name' => 'Cars', 'ID' => '1' }, { 'Name' => 'Fast cars', 'ID' => '2' }],
```
d-Pixie, line 29: I do though strongly question why it hasn't applied the same logic to these???
```
Categories: [{  Name: 'Cars', ID: '1' }, { Name: 'Fast cars', ID: '2' }],
```
That makes a lot more sense to me, in context...

spec/fortnox/api/repositories/examples/find.rb
```diff
@@ -31,13 +31,12 @@
       subject { find_failure }
 
       let(:not_found_id) { '123456789' }
-      let(:find_failure) do
-        when_performing do
+
+      it do
+        expect do
```
d-Pixie, line 36: "It expect do" reads as well as what my 6 year old is writing ...

spec/fortnox/api/repositories/examples/only.rb
```diff
@@ -30,13 +30,11 @@ def repository_only(repository, vcr_cassette, filter)
     end
 
     context 'with invalid filter' do
-      subject do
-        when_performing do
+      it do
+        expect do
```
d-Pixie, line 34: I REALLY don't agree with using a `do ... end` block when you are chaining. That is just bad form. It should at least be:
```
it 'does something here' do
  expect{ repository_only(repository, 'filter_invalid', 'doesntexist') }.to raise_error(Fortnox::API::RemoteServerError, /ogiltigt filter/)
end
```

spec/fortnox/api/repositories/examples/search.rb
```diff
@@ -26,15 +26,13 @@
...
       context 'with special characters' do
-        subject do
-          when_performing do
+        it do
+          expect do
             VCR.use_cassette("#{vcr_dir}/search_with_special_char") do
               repository.search(attribute_hash_key_name => 'special char å')
             end
-          end
+          end.not_to raise_error
```
d-Pixie, line 34: Man, the former style read sooo much better 😢

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -131,15 +131,18 @@ def new_invoice(comments:)
...
 
         let(:updated_persisted_invoice) do
           VCR.use_cassette("#{vcr_dir}/save_old_with_nil_comments") do
             repository.save(persisted_invoice.update(comments: nil))
           end
         end
 
-        pending { is_expected.to eq(nil) }
+        it do
+          pending "test to rerecord VCR cassette, maybe it's working now"
```
d-Pixie, line 143: This shouldn't still be in here, right? Sounds like leftover debug 😄

d-Pixie (changes requested):

fortnox-api.gemspec
```diff
@@ -46,4 +45,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'timecop', '~> 0.9.6'
   spec.add_development_dependency 'vcr', '~> 6.1'
   spec.add_development_dependency 'webmock', '~> 3.18'
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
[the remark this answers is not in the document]
ehannes: I've already fixed MFA :) See Slack.

spec/fortnox/api/mappers/base/from_json_spec.rb
ehannes: I'll fix it. I guess it just didn't catch this...

spec/fortnox/api/repositories/invoice_spec.rb
ehannes: It should still be there. I'll have a look when I rerecord the cassettes.

spec/fortnox/api/repositories/examples/find.rb
```diff
@@ -28,16 +28,13 @@
     end
 
     context 'when not found' do
-      subject { find_failure }
```
ehannes, line 31: This was never used so I just remove it...

fortnox-api.gemspec
```diff
@@ -16,6 +16,9 @@ Gem::Specification.new do |spec|
                        'easily persistable to Fortnox again using the `save` method.'
   spec.homepage      = 'http://github.com/accodeing/fortnox-api'
   spec.license       = 'LGPL-3.0'
+  spec.metadata = {
+    'rubygems_mfa_required' => 'true'
```
d-Pixie, line 20: Doesn't like the newer hash syntax? `rubygems_mfa_required: 'true'`, also should this be a string `'true'` or the singleton `true`?

d-Pixie (approved):

fortnox-api.gemspec
ehannes: Keys must be strings since they are parsed to json. Same with the value, must also be a string: `metadata['rubygems_mfa_required'] value must be a String (Gem::InvalidSpecificationException)`

PR #236, opened 2023-01-30
Rerecord vcr cassettes

ehannes:


lib/fortnox/api/models/base.rb
```diff
@@ -41,6 +41,26 @@ def unique_id
           send(self.class::UNIQUE_ID)
         end
 
+        # This filtering logic could be improved since it is currently O(N*M).
```
ehannes, line 44: Note: This code was incorrectly removed when dry-libs were upgraded.

lib/fortnox/api/repositories/base.rb
```diff
@@ -134,6 +139,16 @@ def renew_access_token
             refresh_token: @token_store.refresh_token
           }
 
+          if Fortnox::API.debugging
```
ehannes, line 142: For debugging token renewal.

spec/fortnox/api/repositories/article_spec.rb
```diff
@@ -27,10 +29,10 @@
                    :ean,
                    '5901234123457'
 
-  # When recording new VCR cassettes, expected matches must be increased
-  include_examples '.all', 1
+  include_examples '.all'
```
ehannes, line 32: These examples are refactored, so no explicit number is required anymore.

spec/support/helpers/configuration_helper.rb
```diff
@@ -3,36 +3,74 @@
 require 'dotenv'
 require 'jwt'
 
-Dotenv.load('.env.test')
+DOTENV_FILE_NAME = '.env.test'
+Dotenv.load(DOTENV_FILE_NAME)
+
+REFRESH_TOKENS = ENV.fetch('REFRESH_TOKENS')
```
ehannes, line 9: Added to support using a real token, for instance when rerecording VCR cassettes as I were doing now.

spec/fortnox/api/repositories/invoice_spec.rb
```diff
@@ -90,12 +107,6 @@ def save_invoice(country:, vcr_cassette: country)
         it { is_expected.to eq('GB') }
       end
 
-      describe 'KR' do
```
ehannes, line 93: Apparently, this was not a valid call anymore. Fortnox complained about it...

ehannes: I checked the VCR cassette and apparently they are still using the old country name: `Korea, Republic of`. When we upgraded the dependencies, we upgraded the `countries` gem, that hopefully has the correct translation of `KR` to `South Korea`, so I guess Fortnox is just slow on such updates...

ehannes: I'll remove the cassette. Since we don't use South Korea we can just ignore this. If anyone needs South Korea, let's fix that then...

lib/fortnox/api/repositories/base.rb
d-Pixie: Remove this now then?

lib/fortnox/api/request_handling.rb
```diff
@@ -18,12 +18,13 @@ def validate_response(response)
...
         if response.headers['content-type'].start_with?('text/html')
           raise Fortnox::API::RemoteServerError,
-                "Fortnox API's response has content type \"text/html\" instead" \
-                'of requested "application/json".' \
+                "Fortnox API's response has content type \"text/html\" instead of requested \"application/json\"." \
                 'This could be due to invalid endpoint or when the API is down. ' \
                 "Body: #{response.body}"
         end
 
+        raise Fortnox::API::RemoteServerError, 'Unauthorized request' if response.code == 401
```
d-Pixie, line 26: Add one for 403 as well.

d-Pixie: And tell them what the actual error is 🙂 "No access token provided" and "Invalid or expired access toekn provided".

spec/fortnox/api/repositories/examples/all.rb
```diff
@@ -1,16 +1,16 @@
...
+RSpec.shared_examples_for '.all' do
   describe '.all' do
     let(:response) do
       VCR.use_cassette("#{vcr_dir}/all") { subject.all }
     end
 
-    specify 'returns correct number of records' do
-      expect(response.size).to be count
+    it 'returns a list of records' do
+      expect(response).not_to be_empty
```
d-Pixie, line 10: Weaker test. Any reason why, other than not having to update the tests after each change? One possibility would be to put all of these in some kind of central config for specs instead and read them from there. That way you could update them all in the same place.

spec/fortnox/api/repositories/invoice_spec.rb
d-Pixie: 👍

lib/fortnox/api/repositories/base.rb
ehannes: yes 👍

spec/fortnox/api/repositories/examples/all.rb
ehannes: Yes, that was the point. But you are correct, this is a weaker test. I'll add back the numbers.

I think I'll stick with the numbers in the specs. You need to update other numbers there as well, so better have all of them in that context.

ehannes:
Build fails due to renew logic in tests, which I will remove in the upcoming PR, so I'll merge this with failing builds...

PR #237, opened 2023-01-31
[WIP] Authentication repository

ehannes:
- Adds Authentication repository

### DIFF
--- a/lib/fortnox/api/repositories/authentication.rb
+++ b/lib/fortnox/api/repositories/authentication.rb
@@ -1,0 +1,76 @@
+require_relative 'base'
+
+module Fortnox
+  module API
+    module Repository
+      class Authentication
+        def renew_access_token(refresh_token)
+          raise ArgumentError, "Refresh token is empty!" if refresh_token.nil? || refresh_token.empty?
+
+          credentials = Base64.encode64("#{client_id}:#{client_secret}")
+
+          renew_headers = {
+            'Content-type' => 'application/x-www-form-urlencoded',
+            Authorization: "Basic #{credentials}"
+          }
+
+          body = {
+            grant_type: 'refresh_token',
+            refresh_token: refresh_token
+          }
+
+          response = HTTParty.post(config.token_url, headers: renew_headers, body: body)
+
+          if response.code == 400
+            raise Fortnox::API::RemoteServerError, "Bad request. Error: \"#{response.body}\""
+          end
+
+          if response.code == 401
+            raise Fortnox::API::RemoteServerError, "Unauthorized request. Error: \"#{response.body}\""
+          end
+
+          if response.code != 200
+            message = 'Unable to renew access token. ' \
+                      "Response code: #{response.code}. " \
+                      "Response message: #{response.message}. " \
+                      "Response body: #{response.body}"
+
+            raise Exception, message
+          end
+          
+          response_to_hash(response.parsed_response)
+        end
+
+        private
+
+        def response_to_hash(parsed_response)
+          {
+            access_token: parsed_response['access_token'],
+            refresh_token: parsed_response['refresh_token'],
+            expires_in: parsed_response['expires_in'],
+            token_type: parsed_response['token_type'],
+            scope: parsed_response['scope'],
+          }
+        end
+
+        def client_id
+          client_id = config.client_id
+          raise MissingConfiguration, 'You have to provide your client id.' unless client_id
+
+          client_id
+        end
+
+        def client_secret
+          client_secret = config.client_secret
+          raise MissingConfiguration, 'You have to provide your client secret.' unless client_secret
+
+          client_secret
+        end
+
+        def config
+          Fortnox::API.config
+        end
+      end
+    end
+  end
+end
--- a/spec/fortnox/api/repositories/authentication_spec.rb
+++ b/spec/fortnox/api/repositories/authentication_spec.rb
@@ -1,0 +1,138 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'fortnox/api'
+require 'fortnox/api/repositories/authentication'
+
+describe Fortnox::API::Repository::Authentication, integration: true, order: :defined do
+  include Helpers::Configuration
+
+  subject(:repository) { described_class.new }
+
+  before { set_api_test_configuration }
+
+  let(:vcr_dir) { 'authentication' }
+
+  describe '#renew_access_token' do
+    context 'with nil' do
+      it 'raises an error' do
+        expect { repository.renew_access_token(nil) }.to raise_error(ArgumentError, 'Refresh token is empty!')
+      end
+    end
+
+    context 'with empty string' do
+      it 'raises an error' do
+        expect { repository.renew_access_token('') }.to raise_error(ArgumentError, 'Refresh token is empty!')
+      end
+    end
+
+    describe 'missing configuration' do
+      let(:calling_renew_access_token) { repository.renew_access_token('access_token')}
+
+      context 'without client_id' do
+        before { Fortnox::API.configure { |c| c.client_secret = 'a secret' } }
+
+        it 'raises an error'  do
+          expect { calling_renew_access_token }.to raise_error(Fortnox::API::MissingConfiguration, /client id/)
+        end
+      end
+
+      context 'without client_secret' do
+        before { Fortnox::API.configure { |c| c.client_id = 'an id' } }
+
+        it 'raises an error'  do
+          expect { calling_renew_access_token }.to raise_error(Fortnox::API::MissingConfiguration, /client secret/)
+        end
+      end
+    end
+
+    context 'with invalid client' do
+      before do
+        Fortnox::API.configure do |config|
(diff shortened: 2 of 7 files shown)


lib/fortnox/api/repositories/authentication.rb
```diff
@@ -0,0 +1,78 @@
...
+            'Content-type' => 'application/x-www-form-urlencoded',
+            Authorization: "Basic #{credentials}"
+          }
+
+          body = {
+            grant_type: 'refresh_token',
+            refresh_token: refresh_token
+          }
+
+          response = HTTParty.post(config.token_url, headers: renew_headers, body: body)
```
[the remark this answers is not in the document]
ehannes: The token_url is set to `'https://apps.fortnox.se/oauth-v1/token'` by default. The ID and secret are used in the headers, not in the URL. Maybe you are thinking about the URL for getting the tokens in the first place?

lib/fortnox/api/repositories/authentication.rb
```diff
@@ -0,0 +1,78 @@
...
+            expires_in: parsed_response['expires_in'],
+            token_type: parsed_response['token_type'],
+            scope: parsed_response['scope']
+          }
+        end
+
+        def handle_errors(response)
+          raise Fortnox::API::RemoteServerError, "Bad request. Error: \"#{response.body}\"" if response.code == 400
+
+          if response.code == 401
```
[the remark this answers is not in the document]
ehannes: Are we? This repository does not inherit from `Repository::Base` since it's so special, so there's no other place we catch `401` here.

Regarding `403`, how would we get such an error when asking for new tokens?

lib/fortnox/api/repositories/authentication.rb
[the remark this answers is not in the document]
d-Pixie: Yeah, probably 🙂

lib/fortnox/api/repositories/authentication.rb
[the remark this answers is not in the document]
d-Pixie: Invalid refresh token => 403.

Ok, but the error handling is the same? But maybe other things would interfere if you inherited from `Repository::Base`?

.rubocop.yml
```diff
@@ -9,7 +9,7 @@ RSpec/FilePath:
   Enabled: false
 
 RSpec/NestedGroups:
-  Max: 4
+  Max: 5
```
d-Pixie, line 12: 😆 Classic solution to a rules violation, just increase what is allowed (not saying it's wrong per se, but maybe we should talk about it before we change the rules in a repo?).

CHANGELOG.md
```diff
@@ -10,7 +10,10 @@ and this project adheres to
 
 ### Added
 
-- Script in `bin/get_tokens` to issue access and refresh tokens.
+- Script in `bin/get_tokens` to issue access and refresh tokens
+- Script in `bin/renew_tokens` (convenience feature for developers)
+- `Fortnox::API::Repository::Authentication.renew_token` to request token
```
d-Pixie, line 15: "Uses", "Use"? Missing a word or two here to make sense ...

README.md
```diff
@@ -205,24 +205,40 @@ Things you need:
...
+The gem exposes a specific repository for renewing tokens. You use it like this:
+
+```ruby
+require 'fortnox/api'
 
-Note that when the gem automatially fetches a new access token, both the client
-id and client secret are required to generate the authorization headers for that
-request. Therefore, the gem needs access to those values as well.
+# Request new tokens
+tokens = Fortnox::API::Repository::Authentication.new.renew_tokens(your_refresh_token)
```
d-Pixie, line 222: Could add:
```
def self.renew_tokens( token )
  self.new.renew_tokens( token )
end
```
That way you wouldn't have to have the `new` here ... But I guess this is how all the repos work so ...

README.md
```diff
@@ -205,24 +205,40 @@ Things you need:
...
+```
+
+It's up to you to provide a valid token to the gem and to renew it regularly,
+otherwise you need to start over again with the
+[Initialization](#initialization).
 
-As described in the [Authorization](#authorization) section, the gem refreshes
-the tokens automatically, but you need to initialize the process. There a script
-for getting tokens. See [Get tokens](#get-tokens).
+Note that when the gem fetches new tokens, both the client id and client secret
```
d-Pixie, line 239: "when the gem fetches" sounds like it is doing it automatically, but we are not doing that anymore, right? So maybe "when you use the gem to fetch" instead? Making it a passive sentence (from the gems pov)

README.md
```diff
@@ -245,124 +261,25 @@ end
 | --------------- | ----------------------------------------------- | -------- | --------------------------------------------------------------- |
 | `client_id`     | The client id from your Fortnox integration     | Yes      | `nil`                                                           |
```
d-Pixie, line 262: It these are only needed when you manually do requests to refetch tokens we should have them as arguments to the methods that need them instead of giving the gem permanent access to something it won't need regularly.

README.md
```diff
@@ -245,124 +261,25 @@ end
...
-the provided store.**
-
 ### Support for multiple Fortnox accounts
 
-Yes, we support working with several accounts at once. Simply provide more than
-one token store, where each store represents a Fortnox account. Each store needs
-its own refresh and access token. If you provide a `:default` token store, this
-is used as default by all repositories.
+Yes, we support working with several accounts at once. Simply switch access
+token between calls.
```
d-Pixie, line 272: 👍🏽

bin/renew_tokens
```diff
@@ -0,0 +1,29 @@
...
+
+require 'bundler/setup'
+require 'fortnox/api'
+require 'dotenv'
+
+DOTENV_FILE_NAME = '.env'
+Dotenv.load(DOTENV_FILE_NAME)
+
+Fortnox::API.configure do |config|
+  config.client_secret = ENV.fetch('FORTNOX_API_CLIENT_SECRET', nil)
```
d-Pixie, line 12: Yeah, this is what I mean ... We should offload this as well to the client application instead of requiring them to configure it via the gem.

lib/fortnox/api/repositories/authentication.rb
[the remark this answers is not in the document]
ehannes: I'll try to send them an invalid token and see what will happen.

lib/fortnox/api/repositories/base.rb
```diff
@@ -9,9 +9,6 @@
 require_relative '../request_handling'
 
 # TODO: Temporarily disables metrics since this will be rewritten soon...
```
d-Pixie, line 11: Well, now that you don't disable the metrics anymore ... the comment could also be removed?

lib/fortnox/api/repositories/authentication.rb
```diff
@@ -0,0 +1,84 @@
...
+
+          return unless response.code != 200
+
+          message = 'Unable to renew access token. ' \
+                    "Response code: #{response.code}. " \
+                    "Response message: #{response.message}. " \
+                    "Response body: #{response.body}"
+
+          raise Exception, message
+        end
```
d-Pixie, line 62:
```suggestion
        def validate_response(response)
          return if response.code == 200
          
          case response.code
          when 400 then
            raise Fortnox::API::RemoteServerError, "Bad request. Error: \"#{response.body}\""
          when 401 then
            raise Fortnox::API::RemoteServerError, "Unauthorized request. Error: \"#{response.body}\""
          else
            message = 'Unable to renew access token. ' \
              "Response code: #{response.code}. " \
              "Response message: #{response.message}. " \
              "Response body: #{response.body}"

            raise Exception, message
          end
        end
```

d-Pixie (changes requested):

lib/fortnox/api/repositories/authentication.rb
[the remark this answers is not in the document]
ehannes: An invalid refresh token is tested here:

```
      before do
        # VCR: Valid config required
          config.client_secret = ENV.fetch('FORTNOX_API_CLIENT_SECRET')
          config.client_id = ENV.fetch('FORTNOX_API_CLIENT_ID')
        end
      end

      context 'with invalid refresh token' do
        it 'raises an error' do
          expect do
            VCR.use_cassette("#{vcr_dir}/invalid_refresh_token") do
              repository.renew_tokens('invalid_refresh_token')
            end
          end.to raise_error(Fortnox::API::RemoteServerError, /Bad request(.)*Error:(.)*Invalid refresh token/)
        end
      end
```

it returns
```
...
status: 400
'{"error":"invalid_grant","error_description":"Invalid refresh token"}'
```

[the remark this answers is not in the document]
ehannes: `Repository::Base` requires an `access_token`, a mapper and `keys_to_filter` etc. We can't use that class here.

README.md
ehannes: Ye, maybe stick to the convention...

d-Pixie (approved):

PR #241, opened 2024-01-22
Patch 0.8.1

ehannes:
Cherry picks commits to add support for Ruby 3 to v0.8


lib/fortnox/api/models/customer.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require 'fortnox/api/types'
-require 'fortnox/api/models/base'
+require_relative 'base'
+require_relative '..//types'
```
d-Pixie, line 4: Double slash not needed, I assume? 🙂

d-Pixie (changes requested):

spec/fortnox/api/repositories/invoice_spec.rb
d-Pixie, line 1: Why are we removing South Korea?

spec/fortnox/api/repositories/unit_spec.rb
```diff
@@ -19,16 +19,16 @@
   # When recording new VCR cassettes, code needs to be changed to a unique value
   include_examples '.save',
                    :description,
-                   additional_attrs: { code: 'blarg5' }
+                   additional_attrs: { code: 'blarg7' }
```
d-Pixie, line 22: ... Could we in future use reasonable strings for test data? Run a lorem word if you can't think of something appropriate 🙂

spec/fortnox/api/repositories/invoice_spec.rb
ehannes: The country code was changed and we didn't really need this test. We already have multiple tests with different country codes.

spec/fortnox/api/repositories/unit_spec.rb
ehannes: Haha, ye sure 😄

d-Pixie (approved):

